PR #21, opened 2021-07-19
fix: wrong Glee version displayed in welcome message

fmvilas:


src/lib/logger.js
```diff
@@ -1,12 +1,16 @@
...
+import { fileURLToPath } from 'url'
 import util from 'util'
 import chalk from 'chalk'
 import emojis from 'emojis'
 import wordWrap from 'word-wrap'
 
 export { chalk }
 
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = dirname(__filename)
```
magicmatatjahu, line 12: Why you use the `import.meta...` not classic `__dirname`?

fmvilas: https://nodejs.org/api/esm.html#esm_no_filename_or_dirname

fmvilas: This is one of the trade-offs of using ES Modules. I hope they come up with a solution soon :/

magicmatatjahu: LOL I always transform ESM to CJS in Node and I don't know about it. Good to know. Thanks!

magicmatatjahu (approved):

PR #30, opened 2021-07-26
refactor: remove constants and introduce configs

jonaslagoni:
**Description**
At the moment we set constants and then load the configs from the file, however, that config file might contain constants that should be shared across files. This PR moves the functionality to be present in the `constants.js` file.

This is important when those configs are later used in the java runtime for example.


src/lib/constants.js
```diff
@@ -6,16 +7,37 @@ let GLEE_FUNCTIONS_DIR
...
   GLEE_FUNCTIONS_DIR = path.resolve(GLEE_DIR, config.functionsDir || 'functions')
   GLEE_CONFIG_FILE_PATH = path.resolve(GLEE_DIR, 'glee.config.js')
   ASYNCAPI_FILE_PATH = path.resolve(GLEE_DIR, 'asyncapi.yaml')
-
+  await loadConfigsFromFile();
+  
   return getConstants()
 }
 
+async function loadConfigsFromFile(config){
```
fmvilas, line 21: Should this be in another file? I honestly would not expect a "load config" in a file called `constants.js`. Maybe move it to a `config.js` file?

jonaslagoni: I see your point, I think this comes down to what is your definition of `constants` and what is the relationship with configs? 🤔 

Should there be both?

fmvilas: Yeah, I think we could have `constants.js` and `config.js`. That's why I was suggesting :)

To me, `constants.js` is about the constants we use in our code. `config.js` is about handling the user configuration.

jonaslagoni: But what happens when I want to use `GLEE_LIFECYCLE_DIR`, which is partly in constants, but also in config 😅 Gonna be a jumbled mess to make sure all inputs are accounted for and imported if we keep using static configurations at least.

To me, there should not be any changes between what can be configured and what is constants. Constants to me mean default values for configurations. However, here we start to mix and match the meaning which to me is confusing 😆 

For clarity, how do you see https://github.com/asyncapi/glee/blob/256dd39b90fc6d027bb39b9ec971a6db3e315d02/src/lib/runtimes/java.js#L43 access the configuration `GLEE_LIFECYCLE_DIR` if they are both "defined" in `constants.js` and `config.js`?

fmvilas: I see. What about renaming `constants.js` to `config.js`. It will include both configurable and non-configurable constants + the function to read the config.

jonaslagoni: Makes sense to me yes, gonna adapt this PR with the change 👍

fmvilas (approved):

PR #126, opened 2021-10-04
Removing shadowing

NuclearCactus:
Issue#115

**Description**

Copied Adapter object into a new object called copyOfAdapter and used it in the adapters array. This will remove variable shadowing.


magicmatatjahu:
@NuclearCactus Some insights:
- please remove not needed code, `runtimes/**/*` content. Probably you generated examples and pushed them.
- as I wrote about problem in mentioned issue:

> Adapter variable in the addAdapter function is needed: the single item/object in the adapters array has exactly Adapter field, so you should propagate the first argument to new item in the adapters array. How to do it I won't say, that's what learning is all about :)

and you saved reference of imported abstract class `Adapter` to the `copyOfAdapter` field and then propagated it to the new item in the `adapters` array.

You should change signature of function to this:

```js
  addAdapter (adapter, { serverName, server, parsedAsyncAPI }) {
    // logic - `adapter` argument is needed
  }
```

and rest you should handle :)

PR #146, opened 2021-10-17
feat: support different asyncapi file extensions (yaml, yml, json)

### LINKED ISSUE #23: Support different asyncapi file names: asyncapi.json and asyncapi.yml
#### Reason/Context
Some people are using the JSON format of AsyncAPI or they name their files as `asyncapi.yml` (notice the `.yml` instead of `.yaml`).

#### Description
It would be awesome if Glee could auto-detect that you're using one of those files. It could also complain when you use many of them at the same time. For instance, if it finds `asyncapi.yaml` and `asyncapi.json`.

panwauu:
**Description**
- This feature allows for different asyncapi file extensions
- Instead of requiring a asyncapi.yaml also a asyncapi.json or asyncapi.yml is allowed
- Throws errors if none or more than one file is found
- Adds a test case (Which needed an update in the packages and some config files)
- Updates the documentation (If you have a better idea than the regex style feel free)

**Related issue(s)**
Fixes #23

### DIFF
--- a/docs/app-structure.md
+++ b/docs/app-structure.md
@@ -3,16 +3,16 @@
 Glee expects your project to have some files and folders with special names. The best way to get started with Glee is using [create-glee-app](https://github.com/asyncapi/create-glee-app), which sets up everything automatically for you.
 
 ```
-├─ functions          (required)
+├─ functions                    (required)
 │  ├─ onHello.js
 │  └─ ...
-├─ lifecycle          (optional)
+├─ lifecycle                    (optional)
 │  ├─ onConnect.js
 │  └─ ...
-├─ .env               (optional)
-├─ asyncapi.yaml      (required)
-├─ glee.config.js     (optional)
-├─ package.json       (required)
+├─ .env                         (optional)
+├─ asyncapi.(yaml|yml|json)     (required)
+├─ glee.config.js               (optional)
+├─ package.json                 (required)
 ```
 
 |File/Directory|Description|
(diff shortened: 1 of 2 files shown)


src/lib/configs.js
```diff
@@ -12,25 +13,44 @@ export async function setConfigs(config) {
...
+  const asyncApiFiles = ['asyncapi.yaml', 'asyncapi.json', 'asyncapi.yml']
+
+  const promiseArray = asyncApiFiles.map(async (filename) => {
+    return access(path.resolve(baseDir, filename), constants.R_OK)
+      .then(() => { return path.resolve(baseDir, filename) })
+  })
+
+  const fileResults = await Promise.allSettled(promiseArray)
+  const accessiblePaths = fileResults.filter((result) => result.status === 'fulfilled')
+  if (accessiblePaths.length === 0) { throw new Error('AsyncApi file was not found or is not readable') }
```
fmvilas, line 35:
```suggestion
  if (accessiblePaths.length === 0) { throw new Error('AsyncAPI file was not found or is not readable') }
```

src/lib/configs.js
```diff
@@ -12,25 +13,44 @@ export async function setConfigs(config) {
...
+  if (accessiblePaths.length > 1) { throw new Error('Multiple AsyncApi files were found - please remove all but one') }
```
fmvilas, line 36:
```suggestion
  if (accessiblePaths.length > 1) { throw new Error('Multiple AsyncAPI files were found - please remove all but one') }
```

Regarding the message, I think we shouldn't advocate for removing files but to encourage devs to configure Glee to pick one of them. Maybe using the glee.config.js file. Happy if you want to give it a try in a different PR or here. As you prefer.

fmvilas (changes requested):
I'm so sorry for the late review 🙏 Left a couple of comments. Happy if you want to tackle them in a separate issue/PR.

fmvilas:
@panwauu do you plan to work on this PR?

PR #154, opened 2021-10-31
feat: read channel.servers before falling back to x-servers

fmvilas:
**Description**
So far, Glee has been relying on the `x-servers` extension to link channels with servers. I'm now defaulting to the 2.2.0 channel servers feature instead of the `x-servers` extension.


jonaslagoni (approved):
Just so it's said, you are making a breaking change from looking after the servers in channels instead of operations 😆 

That said, LGTM 👍

fmvilas:
Yeah, that's fine. We haven't reached v1 yet. It's expected to happen.

PR #212, opened 2021-12-23
feat: migrate to typescript round 1

fmvilas:
**Description**

- Start porting Glee to TypeScript
- Allows the user to use TypeScript in their projects
- Removes the existing "runtimes" mechanism. We only allow TS/JS. As suggested in #207.

**Related issue(s)**
#144

PR #213, opened 2021-12-24
feat: migrate to typescript (round 2)

### LINKED ISSUE #144: Port Glee to TypeScript
#### Reason/Context
As the spec already defines all the types and schema regarding the communication, Glee misses out on a lot of benefits that typescript could provide here. After all typescript was also written with a smiling developer in mind 😄.

#### Description
Porting to typescript will require large changes in the codebase. Each js file will have to be touched and also the execution process has to be altered (Either using ts-node or typescripts tsc). Since no tests exist it is going to be hard to ensure the behavior does not change and no bugs are introduced.

#### Approach
?

fmvilas:
#212 Follow up

Do not review until #212 is merged 🙏 

Fixes #144


.eslintignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 runtimes
```
jonaslagoni, line 2:
```suggestion
```

jonaslagoni (changes requested):
There is a couple of files under `runtimes/java` that needs to be removed.

And just one small suggestion, otherwise it looks great!

jonaslagoni (approved):

src/lib/logger.ts
```diff
@@ -5,24 +5,44 @@ import util from 'util'
...
 const __dirname = dirname(__filename)
 const TSLogo = chalk.bgHex('#3178C6').white(' TS')
 
-const highlightWords = (words, text) => {
+const highlightWords = (words: string[], text: string) => {
   let result = text
 
   words.filter(Boolean).forEach(word => {
-    result = result.replace(new RegExp(word, 'g'), chalk.white(word))
+    result = result.replaceAll(word, chalk.white(word))
```
jonaslagoni, line 39: This might be a problem, as it is [only supported in node > v15](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll).

jonaslagoni (approved):
One comment, not sure how relevant though.

src/lib/logger.ts
fmvilas: Fixed 👍 Good catch!

jonaslagoni (approved):

PR #230, opened 2022-01-11
feat: add tests

### LINKED ISSUE #149: Enable test in github actions
In #146 I introduced the first test and if it gets approved I plan to add them whenever I implement a new feature and also add them for existing code. 
If the PR with the test gets merged it would be nice to include test in the CI.

fmvilas:
Resolves #149


fmvilas:
I've tried with this great suggestion from @kograte but can't make it work because I'm import'ing modules appending the `.js` extension:

Captura de pantalla 2022-01-11 a las 12 27 12

Which now makes me think that probably we can get rid of these `.js` extensions somehow? 🤔

fmvilas:
Pinging my TS peeps @jonaslagoni @magicmatatjahu

jonaslagoni:
You can remove all the extension yes 🙂 - https://www.typescriptlang.org/docs/handbook/module-resolution.html#classic

fmvilas:
> Without changing anything else 

@jonaslagoni you wish 😄 I can't currently use the code as it is because I removed `"type": "module"` from package.json, therefore I can't use `import`. If I bring it back, tests don't work 😅 I tried compiling down to ES5 and CommonJS modules but then I have a problem trying to require code that's using `import` like the one in the `functions` folder of the user code.

What a fun time I'm having here 😂

fmvilas:
Did you try running an example of the examples folder with these changes?

jonaslagoni:
> Did you try running an example of the examples folder with these changes?

No. Tested them now, and yea they don't work 😛 

And, it's probably because the target environment is wrong. [Node always treat module as CommonJS modules](https://nodejs.org/api/esm.html#enabling). I have not tried to change that default behavior before so if you want you can try and make that work 😄 

I think it is easier to change the target to CommonJS modules though, i.e. the root tsconfig.json file would need to change to:
```json
{
  "compilerOptions": {
    "outDir": "./dist",
    "allowJs": false,
    "target": "es6",
    "esModuleInterop": true,
    "moduleResolution": "node",
    "module": "CommonJS",
  },
  "include": [
    "./src/**/*"
  ],
  "exclude": ["node_modules", "lib", "examples"]
}
```
The next problem for the examples is that  (at least on windows this happens), the CLI is not installed under `./node_modules/.bin/` making it impossible to execute it through the `glee` command. I.e. I changed the run script to `"dev": "node ./node_modules/@asyncapi/glee/dist/cli/index.js dev"` as I haven't researched how to fix that problem.

Next is the transpilation of your example code, you need to adapt the compiler process to use the following compiler options: 
```json
      compilerOptions: {
        allowJs: true,
        target: 'es6',
        esModuleInterop: true,
        moduleResolution: 'node',
        module: "CommonJS",
      }
```
and remove the `"type": "module",` from the example `package.json` file.

Afterward, everything works fine. At least with dummy example 😆 

And the tests are passing fine as well.

This should give you a good basis for continuing 🙂

jonaslagoni:
Instead of all that gibberish, maybe this is enough? https://kulshekhar.github.io/ts-jest/docs/guides/esm-support 🤔

fmvilas:
Yay! That works @jonaslagoni 🎉 Thanks a lot! 🙏

fmvilas:
Gonna stop here. I think they're enough and, if I continue, the PR will be huge.

fmvilas:
Created a follow up issue: #235

jonaslagoni (approved):
LGTM 👍 

Optionally you could add something like this https://github.com/asyncapi/modelina/blob/master/.github/workflows/coverall.yml

jonaslagoni (approved):

PR #234, opened 2022-01-16
feat: made message sending api dry

### LINKED ISSUE #192: feat: Glee's message sending API should be as DRY as possible
Glee supports message handlers (which are invoked upon receiving a message) as well as the sending of messages (typically from within message handlers).

- When sending a message, a Glee app makes use of `OutboundMessage`s, which can be used to `send`, `reply`, and/or `broadcast` a message (https://github.com/asyncapi/glee/blob/master/docs/functions.md).
- Also, `OutboundMessage` contains the `channel` property, which specifies the channel the message is sent to.
- Lastly, the message interactions performed by the Glee app are specified by an AsyncAPI document.

All the above options (`send`, `reply`, `broadcast` and `channel`) specify various aspects of sending a message, which must however always be in accordance with the AsyncAPI document of the Glee application. There is significant overlap between the information contained in these options. In other words, the current API o send messages is not DRY.

Ideally, given an AsyncAPI document, a Glee app should just need one operation to send a message to a channel. Specifically:

- The differences between `send` and `broadcast` should not matter to the Glee app.
- A `reply` is just a send to a reply channel specified by the sender of the incoming message, optionally with the message ID of the incoming message included in the reply message. All of that information should already be in the AsyncAPI document.

In general, every aspect of the messages sent by the Glee app should be specified in the AsyncAPI document - if at all possible - and not be repeated in Glee's message sending API. Only information that cannot be expressed in AsyncAPI should be part of Glee's message sending API (or be included in future versions of the AsyncAPI spec ;) ).

See https://asyncapi.slack.com/archives/CQVJXFNQL/p1637667052282600 for the origins of this request.

sudoshreyansh:
**Description**

- Removed broadcast option while sending messages. Only send and reply are available, as mentioned [here](https://github.com/asyncapi/glee/issues/192#issuecomment-978165400).

**Related issue(s)**
Resolves #192

### DIFF
--- a/examples/dummy/functions/onUserSignedUp.ts
+++ b/examples/dummy/functions/onUserSignedUp.ts
@@ -2,7 +2,7 @@ export default async function (event) {
   const user: any = event.payload
   console.log(`${user.displayName} has recently signed up. Sending an email to ${user.email}.`)
   return {
-    broadcast: [{
+    send: [{
       server: 'websockets',
       payload: event.payload,
     }]
--- a/src/lib/functions.ts
+++ b/src/lib/functions.ts
@@ -60,6 +60,7 @@ export async function trigger({
           headers: msg.headers,
           channel: msg.channel || message.channel,
           serverName: msg.server,
+          broadcast: true,
         }))
       })
     }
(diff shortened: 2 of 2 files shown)


fmvilas (approved):
LGTM 👍 

@GeraldLoeffler since you created the issue you may want to have a look too.

fmvilas:
Merging this one. @GeraldLoeffler please feel free to leave a comment afterward and we can happily address it on a future PR. Thanks, @sudoshreyansh!

fmvilas:
@derberg looks like `/rtm` is not removing the `do-not-merge` label?

derberg:
@fmvilas it should not, they it would not make sense if `/rtm` could merge even if `do-not-merge` is there. You can only remove it manually as a codeowner with write access

fmvilas:
Oh, you're right 🤦

PR #240, opened 2022-01-22
feat: clusterization

### LINKED ISSUE #26: Add support for clusterization
#### Reason/Context
Currently, Glee would not behave great when spanning multiple instances. The reason is the [`broadcast`](https://github.com/asyncapi/glee/blob/master/docs/functions.md) option. As it is right now, we're storing connections in memory, so when a function responds saying it wants to broadcast a message, it goes through all the connections and sends the message. Obviously, if we have multiple instances, the message will be "broadcasted" only to a subset of clients (the ones that a given instance has in memory).

Relevant parts of the code are:

* https://github.com/asyncapi/glee/blob/master/src/lib/adapter.js#L23
* https://github.com/asyncapi/glee/blob/master/src/lib/adapter.js#L69
* https://github.com/asyncapi/glee/blob/master/src/lib/adapter.js#L82
* https://github.com/asyncapi/glee/blob/master/src/lib/adapter.js#L36

#### Description
We should allow for clusterization of the multiple instances. There are multiple approaches for this but I think the simplest one would be to keep things in memory and let the user choose a mechanism to communicate the different instances, e.g., Redis.

It would be something like:

1. A message with the broadcast option enabled is requested to be sent.
2. The adapter sends a message to Redis saying it wants to broadcast a message.
3. All the instances receive the message and proceed to send it.

The advantage of this approach is that it's simple but it doesn't guarantee all the messages will be sent. For instance, what happens if one of the instances doesn't receive the Redis message? That's something to think about. There are many existing techniques, we don't need to reinvent the wheel.

sudoshreyansh:
**Description**

- Added support for clusterization
- Added relevant tests and documentation

**Related issue(s)**
Fixes #26

### DIFF
--- a/docs/config-file.md
+++ b/docs/config-file.md
@@ -21,13 +21,23 @@ export default async function () {
       httpServer: customServer, // A custom HTTP server of your own.
       adapter: 'native', // Default. Can also be 'socket.io' or a reference to a custom adapter.
       port: process.env.PORT,
+    },
+    cluster: {
+      adapter: 'redis',
+      name: 'cluster', // Default. Name of your cluster.
+      url: 'redis://localhost:6379', // Server URL used by adapter for clustering
     }
   }
 }
 ```
 
-|Field|Description|
-|---|---|
-|adapter|The Glee adapter to use for the WebSocket server. Defaults to a "native" WebSocket implementation. Other allowed values are `socket.io` (to use the [Socket.IO](https://socket.io/) Glee adapter) or a reference to a custom adapter.
-|httpServer|A custom HTTP server of your own. E.g., an [Express](https://expressjs.com/en/4x/api.html) server or any object that implements the [http.Server](https://nodejs.org/api/http.html#http_class_http_server) interface.
-|port|The port to use when binding the WebSocket server. This is useful when your server is behind a proxy and the port exposed for consumption is not the same as the port your application should be bound to. Defaults to the port specified in the selected AsyncAPI server.
+|Parent|Field|Description|
+|---|---|---|
+|websocket|
+||adapter|The Glee adapter to use for the WebSocket server. Defaults to a "native" WebSocket implementation. Other allowed values are `socket.io` (to use the [Socket.IO](https://socket.io/) Glee adapter) or a reference to a custom adapter.
+||httpServer|A custom HTTP server of your own. E.g., an [Express](https://expressjs.com/en/4x/api.html) server or any object that implements the [http.Server](https://nodejs.org/api/http.html#http_class_http_server) interface.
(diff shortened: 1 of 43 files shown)


docs/config-file.md
```diff
@@ -21,13 +21,23 @@ export default async function () {
...
   }
 }
 ```
 
-|Field|Description|
-|---|---|
-|adapter|The Glee adapter to use for the WebSocket server. Defaults to a "native" WebSocket implementation. Other allowed values are `socket.io` (to use the [Socket.IO](https://socket.io/) Glee adapter) or a reference to a custom adapter.
-|httpServer|A custom HTTP server of your own. E.g., an [Express](https://expressjs.com/en/4x/api.html) server or any object that implements the [http.Server](https://nodejs.org/api/http.html#http_class_http_server) interface.
-|port|The port to use when binding the WebSocket server. This is useful when your server is behind a proxy and the port exposed for consumption is not the same as the port your application should be bound to. Defaults to the port specified in the selected AsyncAPI server.
+|Parent|Field|Description|
```
fmvilas, line 34: You don't need this `Parent` column. Maybe you want to follow the dot notation as we're doing here: https://github.com/asyncapi/generator/blob/master/docs/api.md#new-generatortemplatename-targetdir-options.

So field names would be `websocket.adapter` instead of `adapter` and so on.

examples/dummy/glee.config.js
```diff
@@ -1,10 +1,15 @@
...
-  // return {
-  //   websocket: {
-  //     httpServer: customServer,
-  //     adapter: 'socket.io', // Default. Can also be 'socket.io' or a reference to a custom adapter.
-  //   }
-  // }
+  return {
+    // websocket: {
+    //   httpServer: customServer,
+    //   adapter: 'socket.io', // Default. Can also be 'socket.io' or a reference to a custom adapter.
```
fmvilas, line 7: Oh! Just caught an error that you didn't introduce but you may want to fix :)

```suggestion
    //   adapter: 'native', // Default. Can also be 'socket.io' or a reference to a custom adapter.
```

src/lib/message.ts
```diff
@@ -56,6 +60,7 @@ class GleeMessage extends EventEmitter {
     if (serverName) this._serverName = serverName
     if (connection) this._connection = connection
     if (broadcast) this._broadcast = !!broadcast
+    if (cluster) this._cluster = !!cluster
```
fmvilas, line 63: I think it's safe to do this:

```suggestion
    if (cluster) this._cluster = cluster
```

Since TypeScript is taking care that this is always a boolean.

fmvilas (changes requested):
Awesome stuff @sudoshreyansh 🙌 

I left some minor comments.

sudoshreyansh:
@fmvilas I have made the changes.

fmvilas (approved):
👏 Really good and elegant solution, @sudoshreyansh. Great stuff! 🙌

Issue #25, opened 2021-07-19
Add support for an HTTP/FaaS runtime


fmvilas:
#### Reason/Context
We're currently hosting the functions within the same repository but it could be useful to have them as serverless functions.

#### Description
Implement an HTTP runtime that will make an HTTP call to a FaaS (Function as a Service) provider and waits for the HTTP response.

The HTTP response may vary from one provider to another. From the top of my head, we should support AWS Lambda, Netlify functions, Vercel, Azure Functions, GCP Cloud Functions.

fmvilas:
Yeah! The last point is optional though. Depending on the protocol we may not be able to give any feedback to the user.

sudoshreyansh:
@fmvilas The runtime mechanism was removed in #212, so is this feature still relevant?

darp99:
Hello sir, @fmvilas 
This sounds really interesting if no one is working on this issue then i like to work on this for a part of mentorship program but i have to first learn required skills for this so can i get help form you for move ahead.

fmvilas:
Awesome! Yeah, no worries. During the mentorship program, I'll be here to support you. If this issue gets selected, of course 😄

fmvilas:
This is still relevant, @github-actions. Please don't close it 🙏

KhudaDad414:
@fmvilas 
There are a few points that I would need some input on:

1) Do users need to have the FaaS function already in place, or will they be able to write their function in Glee and have us generate the functions for them to upload to their FaaS provider?
2) What should the user interface for this feature look like? I understand that users need to provide the link to their endpoint and the method for calling it, but I'm not sure how they would do this in the interface. Can you provide more details on how this feature would work in practice?

@Souvikns anything from your side?

KhudaDad414:
After considering different approaches to resolving this issue with the help of @fmvilas, three potential solutions were identified.

## 1) Utilizing Extensions in the AsyncAPI File
This solution appears to be an obvious first choice. By adding an extension such as `x-glee-trigger` to the messages or channels in the specification file, Glee can handle it automatically. However, one drawback of this approach is that it would require adding another configuration to specify the TOKEN or similar information in the Glee configuration.

## 2) Exporting Configuration from the Functions

The current approach that we use for lifecycle methods. Where a function and the corresponding triggering event are returned. 
[...]

PR #265, opened 2022-03-07
feat: add http adapter 

Souvikns:
**Description**

Learning the codebase by creating an HTTP adapter. 

Aims to - 
- connect HTTP server as per the specification 
- Handle incoming messages only on the channels specified in the specification 
- Send appropriate responses from the registered function. 

See also #25


src/adapters/http/index.ts
```diff
@@ -0,0 +1,109 @@
...
+                        const {isValid, humanReadableError, errors} = validateData(searchParams, query);
+                        if(!isValid){
+                            const err = new GleeError({humanReadableError, errors});
+                            console.log('ERROR', err);
+                            this.emit('error', err);
+                            res.end('HTTP/1.1 400 Bad Request\r\n\r\n');
+                            return;
+                        }
+                    }
+                }
```
Souvikns, line 71: I am trying to validate the query parameters but getting an issue. All the query parameters are being treated as a string so if my query defined in `asyncapi.yaml` as a string then whatever the `query` is validation results in positive. 

```
bindings:
      http:
        bindingVersion: 0.1.0
        type: request
        method: GET
        query:
          type: object
          properties:
            name:
              type: string

```

for this `?name=shreak`, `?name` and `?name=12` get validated positively which is something that shouldn't happen.

fmvilas: 🤔  I think you have to try to find the type on purpose. It's true that query parameters are always treated as strings so you'll only get positives in this case. You can probably do this: try `Number(query.name)` and if you don't get `NaN` it means it's a number so you can safely do `query.name = Number(query.name)`. Same with `null`. If you get that `query.name` is `undefined` or `null` then you can safely assign it to `null`.

That said, you don't check for `query.name` but for each key in the object. Don't forget about array values. Hope that helps :)

Issue #259, opened 2022-02-24
Add support for WebSocket clients


fmvilas:
#### Reason/Context

Currently, Glee only supports creating a WebSocket server. That makes it unusable when someone wants to build a WebSocket client to connect to an existing WS server.

#### Description

We already have support for WebSocket but it's always assuming we want to create a server, and that's plain wrong. Instead, Glee should default to creating a WS client and only create a server when we mark the `server` definition in the AsyncAPI file as "local".

For instance:

```yaml
asyncapi: 2.3.0
servers:
  myWebsiteWebSocketServer:
    url: 'ws://mywebsite.com/ws'
    protocol: ws
    x-kind: local
  someProductWSServer:
    url: 'ws://someproduct.com/ws'
    protocol: ws
```

The server `myWebsiteWebSocketServer` is a WS server we want to create with Glee but the server `someProductWSServer` is a remote server we want to send/receive information to/from. The former is what's already implemented and the latter is what we should be implementing as part of this issue. Notice the `x-kind: local` on the `myWebsiteWebSocketServer` server. Only when this is found, we should create a server, otherwise, we should default to a client. Another allowed value for `x-kind` is `remote`.

#### Notes

`x-kind` is an extension we're creating as part of this issue, this is not defined anywhere yet so we have some freedom defining how we want it to be. I'm trying to make it as similar as possible to the [Remotes and local servers issue on the spec](https://github.com/asyncapi/spec/issues/689). In v3, we may end up having two root keys: `servers` (local servers) and `remotes` (remote servers or brokers). For v2.x (the current version), I think it's just fine if we add the `x-kind` extension but I'm happy to listen for alternatives.

deepanshu44:
Hello @fmvilas , I just want to ask that can I work on this issue? Is it difficult to solve this issue?
Thanks

fmvilas:
Hey @deepanshu44! I'd say just go ahead but this issue is marked to be part of GSoC (Google Summer of Code) so I'm not sure how to proceed here. @derberg you are our GSoC expert, is it a problem? @deepanshu44 do you plan to participate at GSoC this year?

derberg:
it is definitely ok to remove the label and move under https://www.asyncapi.com/blog/openforce-2022 and if you complete it in March we can at least send you an AsyncAPI t-shirt

derberg:
just remember to update mentioned blog post with link to this issue if you want to commit to Open Force

deepanshu44:
[...]

PR #267, opened 2022-03-08
feat:Add support for WebSocket clients

deepanshu44:
resolves issue [#259](https://github.com/asyncapi/glee/issues/259)


deepanshu44:
@fmvilas I also have a question, 
I cannot understand when the connection to remote is established. 
I mean Glee lib has a [connect](https://github.com/asyncapi/glee/blob/master/src/lib/glee.ts#L123) method, but at what point it is executed and when?

Souvikns:
I think when you emit the `connect` event in the adapter https://github.com/asyncapi/glee/blob/2b2fb5455c034f73851afe9781d73c9116084155/src/adapters/ws/index.ts#L106 is when the `connect` function in `glee.ts` invoked. 
https://github.com/asyncapi/glee/blob/2b2fb5455c034f73851afe9781d73c9116084155/src/lib/glee.ts#L123

> I am new to the project so I might be wrong

fmvilas:
It's called here: https://github.com/asyncapi/glee/blob/master/src/index.ts#L169. Notice the `connect` method on Glee has a `listen` alias. This lib used to be publicly exposed, therefore the alias.

derberg:
@deepanshu44 do you plan to continue working on it? Open Force ended but since you started during the event, it is ok if you complete work later

deepanshu44:
@derberg I am really sorry about this. I will get this done in this week.
I am really sorry

derberg:
@deepanshu44 no need to be sorry. I'm just making sure all is fine. I know many of our contributors, especially those coming from initiatives like Open Force, are mostly students and have their studies-related duties. Cheers 🍺

src/adapters/ws-server/index.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  async send(message: GleeMessage): Promise<void> {
+    return this._send(message)
+  }
+
+  _connect(): Promise<this> {
+    return new Promise((resolve, reject) => {
+      const serverUrl = new URL(this.serverUrlExpanded)
+      const wsHttpServer = this.glee.options?.websocket?.httpServer || http.createServer()
```
Souvikns, line 25: Since we are creating ws client I think we should do something like this 
```suggestion
      const wsClient = new Websocket(serverUrl);
```

fmvilas: This is the WS server but you're right, we should be doing this on the client, which is in the `ws-client-/indext.ts` file.

Souvikns: Actually, I was not able to comment on the `ws-client/index.ts` file. Sorry for any confusion.

deepanshu44:
@fmvilas I have a very small doubt, why we want to create Websocket-client, as, as far as I know, we can create Websocket-client using asyncapi generator?

fmvilas:
Generator and Glee serve two different purposes:

1. Generator takes an AsyncAPI file and generates code from it. Glee doesn't generate anything. Glee **is** a client or server that is aware of the AsyncAPI file.
2. Generator is executed only during development or CI. Glee is executed during development and production. It's a framework that uses AsyncAPI as a config file.
3. With Generator's generated code, if someone touches the AsyncAPI file but doesn't execute Generator again, the code gets outdated and what's described in the AsyncAPI file might not match the code. The opposite is also very common. People update the code but don't update the AsyncAPI file, making the AsyncAPI file outdated. With Glee, this is impossible. If something is not defined in the AsyncAPI file, it can't be used in the code. If something is in the code but not in the AsyncAPI file, Glee will complain that it's not defined in the AsyncAPI file. That makes the code and the AsyncAPI file stay in sync forever. Pretty much like GraphQL does with the code and the `schema.graphql` file.

Does that make sense?

fmvilas:
Be patient, robotito. We're working on it.

fmvilas:
Closing this one in favor of #319.

PR #272, opened 2022-03-14
fix: show correct log for outbound message

### LINKED ISSUE #269: Logs show "broadcasted" when sending to a broker
#### Describe the bug
The dev logs show the "broadcasted" word when sending to a broker. That's incorrect and can lead to confusion.

#### How to Reproduce
Captura de pantalla 2022-03-14 a las 7 58 45

#### Expected behavior
It should show the word "sent". For instance, "post/liked was sent to mosquitto".

However, when we're sending to a WebSockets server, it should say "was broadcasted" unless it was a reply to a specific WebSockets client.

#### Hints
This is the middleware in charge of logging these things: https://github.com/asyncapi/glee/blob/master/src/middlewares/logger.ts.

Souvikns:
**Description**

This PR aims to change the `outBoundMessage` logs 

Now, `broadcast` is logged when the `send` event is triggered for a `websockets` server.

image

**Related issue(s)**

Fixes #269


examples/social-network/websocket-server/functions/onLikeDislike.js
```diff
@@ -33,6 +33,18 @@ export default async function (event) {
...
+        channel: '/',
+        payload: {
+          type: 'likes_count_updated',
+          data: {
+            postId,
+            totalCount
+          }
+        }
+      }
+    ],
```
Souvikns, line 47: @fmvilas when I trigger `reply` event I still get `['send']` from `message.eventName()`. 
Shouldn't it be `['reply']` when `message.reply` function is called. 
https://github.com/asyncapi/glee/blob/2b2fb5455c034f73851afe9781d73c9116084155/src/lib/functions.ts#L72

examples/social-network/websocket-server/functions/onLikeDislike.js
```diff
@@ -33,6 +33,18 @@ export default async function (event) {
         },
       }
     }],
+    reply: [
```
fmvilas, line 36: This one shouldn't be changed to `reply` but actually, the other one above (`broadcast`) should be replaced as `send`. Watch out you'll end up with an object with 2 `send` keys so instead, you should combine the two.

examples/social-network/websocket-server/functions/onLikeDislike.js
fmvilas: Where do you get `['send']` from `message.eventName()`?

src/lib/logger.ts
```diff
@@ -113,8 +113,14 @@ export const logInboundMessage = (message: GleeMessage) => {
 }
 
 export const logOutboundMessage = (message: GleeMessage) => {
+  let verb: string
+  const event = message.eventNames()[0]
   const icon = message.broadcast ? '⇶' : '↗'
-  const verb = message.broadcast ? 'broadcasted' : 'sent'
+  if (message.serverName === 'websockets' && event !== 'reply') {
```
fmvilas, line 119: You can't do `message.serverName === 'websockets'` here because `websockets` is a name the user gave to this server but it doesn't have to be like this. We may call a websockets server `production` or `testing` or even other languages like Spanish or French. So we can't rely on this.

fmvilas (changes requested):

examples/social-network/websocket-server/functions/onLikeDislike.js
Souvikns: yeah in `message.eventName()`, I am trying to differentiate between two different events but I am getting the same parameter.

Souvikns: Oh sorry I mean `message.eventNames()` and I am accessing it in `message` object in the `logOutBoundMessage`. 

> However, when we're sending to a WebSockets server, it should say "was broadcasted" unless it was a reply to a specific WebSockets client. 

For this, I am trying to access the event type of the messages, Now `broadcast` is deprecated and `send` event has `broadcast:true` so I guess we can check the event that way.

src/lib/logger.ts
Souvikns: > However, when we're sending to a WebSockets server, it should say "was broadcasted" unless it was a reply to a specific WebSockets client.

Does this mean logs should have `was broadcasted` for all bidirectional protocols?

examples/social-network/websocket-server/functions/onLikeDislike.js
fmvilas: > Oh sorry I mean message.eventNames()

But where is `message.eventNames()` defined? 🤔 I don't think it's something from our code, is it? I think what you're using is this [`eventNames`](https://nodejs.org/api/events.html#emittereventnames) which I don't think will be useful at all for you. This is something from Node.js EventEmitter so I honestly have no idea what value you'll get there.

src/lib/logger.ts
fmvilas: Actually no. MQTT is also bidirectional and it shouldn't say `was broadcasted`. I think it should rely on the fact that it's a server and not a client. So far, when we "see" `ws` on the spec file, we create a WebSockets **server**. But when we see `mqtt` we create an MQTT **client**. So, I don't think this can be solved before https://github.com/asyncapi/glee/issues/259.

Souvikns: Then I will wait for https://github.com/asyncapi/glee/issues/259 for this one.

fmvilas:
Removing stale label as we're waiting for #259 before continuing on this one.

Souvikns:
resuming work on this.

src/lib/functions.ts
```diff
@@ -105,12 +107,13 @@ export async function trigger({
     }
 
     res?.send?.forEach((msg) => {
+      const isBroadcast = parsedAsyncAPI.server(msg.server).protocol() === 'ws' && !isRemoteServer(parsedAsyncAPI, msg.server)
```
Souvikns, line 110: Only WebSocket and HTTP could be servers, and since currently, we are only supporting WebSocket, thus I am just checking if the message is being sent to a non-remote WebSocket server.

fmvilas (changes requested):
Can you please review your PR first? It has a looot of changes that are not related to logging outbound messages.

src/lib/functions.ts
fmvilas: What about this instead?

```suggestion
      const localServerProtocols = ['ws', 'wss', 'http', 'https']
      const serverProtocol = parsedAsyncAPI.server(msg.server).protocol().toLowerCase()
      const isBroadcast = localServerProtocols.includes(serverProtocol) && !isRemoteServer(parsedAsyncAPI, msg.server)
```

src/lib/util.ts
```diff
@@ -116,3 +117,9 @@ export const gleeMessageToFunctionEvent = (message: GleeMessage, glee:Glee): Gle
     glee,
   } as GleeFunctionEvent
 }
+
+export const isRemoteServer = (parsedAsyncAPI: AsyncAPIDocument, serverName: string): boolean => {
+  const remoteServer = parsedAsyncAPI.extension('x-remoteServers')
+  if (remoteServer && remoteServer.includes(serverName)) return true
+  return false
```
fmvilas, line 124: You can simplify this further:

```suggestion
  const remoteServers = parsedAsyncAPI.extension('x-remoteServers')
  return remoteServers && remoteServers.includes(serverName)
```

Also, renamed the variable to a plural name since it's returning an array of servers, not just one.

fmvilas (changes requested):

fmvilas (approved):

PR #273, opened 2022-03-15
fix: added log for invalid returned data from functions

### LINKED ISSUE #270: Nothing fails/logs when a Glee Function returns an unknown structure
#### Describe the bug
Glee Functions should either return `undefined` or a well-known object containing either `send` or `reply` keys. If I return something different, say an Array —for instance—, Glee doesn't complain in any way and therefore it's hard to debug what's wrong in your code.

#### How to Reproduce
Go change the [returned value here](https://github.com/asyncapi/glee/blob/master/examples/dummy/functions/onUserSignedUp.ts#L4) to an array and see nothing happens. The code will not work but it also doesn't complain.

#### Expected behavior
It should either fail or log a warning telling us the returned structure is not expected and that we should return either `undefined` or the [well-known structure](https://github.com/asyncapi/glee/blob/master/docs/functions.md).

sudoshreyansh:
**Description**

- added check and log for invalid data returned from functions

**Related issue(s)**
Fixes #270

### DIFF
--- a/src/lib/functions.ts
+++ b/src/lib/functions.ts
@@ -54,10 +54,10 @@ export async function trigger({
 }) {
   try {
     const res = await functions.get(operationId).run(gleeMessageToFunctionEvent(message, app))
-    let value = res === undefined
+    let valid = res === undefined
 
     if (res?.send) {
-      value = true
+      valid = true
       res.send.forEach((msg) => {
         app.send(new GleeMessage({
           payload: msg.payload,
@@ -70,7 +70,7 @@ export async function trigger({
     }
 
     if (res?.reply) {
-      value = true
+      valid = true
       res.reply.forEach((msg) => {
         message.reply({
           payload: msg.payload,
@@ -80,7 +80,7 @@ export async function trigger({
       })
     }
 
-    if ( !value ) {
+    if ( !valid ) {
       logWarningMessage(`Function ${operationId} returned invalid data`, {
         highlightedWords: [operationId]
       })
--- a/src/lib/functions.ts
+++ b/src/lib/functions.ts
@@ -54,8 +54,10 @@ export async function trigger({
 }) {
   try {
     const res = await functions.get(operationId).run(gleeMessageToFunctionEvent(message, app))
+    let value = res === undefined
 
     if (res?.send) {
+      value = true
       res.send.forEach((msg) => {
         app.send(new GleeMessage({
           payload: msg.payload,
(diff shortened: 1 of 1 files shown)


fmvilas (changes requested):
I think we should do better here instead of just checking for the existence of `send` or `reply`. What if I return an object like this:

```js
return {
  send: 10,
}
```

It may be a good idea to use JSON Schema for this and we already have `ajv` installed as a dependency :)

sudoshreyansh:
Ok I have made the changes.

@fmvilas do you want to log the errors too while validating the returned value or just the invalid message?

src/lib/functions.ts
```diff
@@ -6,13 +6,46 @@ import { logWarningMessage } from './logger.js'
...
 import { pathToFileURL } from 'url'
 
 interface FunctionInfo {
   run: GleeFunction,
 }
 
+const OutboundMessageSchema = {
+  type: 'object',
+  properties: {
+    payload: { type: ['string', 'object'] },
```
fmvilas, line 19: Payload can be anything so I'd just make it like this:

```suggestion
    payload: {},
```

src/lib/functions.ts
```diff
@@ -6,13 +6,46 @@ import { logWarningMessage } from './logger.js'
...
+    headers: {
+      type: 'object',
+      propertyNames: { type: 'string' },
+      additionProperties: { type: 'string' }
```
fmvilas, line 23:
```suggestion
      additionalProperties: { type: 'string' }
```

src/lib/functions.ts
```diff
@@ -54,10 +87,17 @@ export async function trigger({
 }) {
   try {
     const res = await functions.get(operationId).run(gleeMessageToFunctionEvent(message, app))
-    let valid = res === undefined
+    const { isValid } = validateData(res, FunctionReturnSchema)
+
+    if ( !isValid ) {
+      logWarningMessage(`Function ${operationId} returned invalid data`, {
+        highlightedWords: [operationId]
+      })
```
fmvilas, line 95:
```suggestion
      const err = new GleeError({
        humanReadableError,
        errors,
      })
      err.message = `Function ${operationId} returned invalid data.`
      logError(err, {
        highlightedWords: [operationId]
      })
      
      
```

You probably have to import GleeError above. If I didn't commit any mistake 😅  that should give a better hint to the user on what actually went wrong.

fmvilas (changes requested):

sudoshreyansh:
@fmvilas I have made the changes.

fmvilas (approved):
Great stuff 👏

PR #288, opened 2022-04-25
docs: fix links in Markdown files

KhudaDad414:
#### Description
this PR is supposed to fix all the [broken links](https://github.com/asyncapi/go-watermill-template/runs/6143307572?check_suite_focus=true) in markdown files across the repo.

- ignoring https://github.com/asyncapi/glee/blob/086c449/src since there are lots of links to this location and checking these links causes GitHub to give 429 (Too many requests) error.
- ignoring https://asyncapi.org/glee/docs for now. I guess we are going to have the docs on the website in the future?

### DIFF
--- a/mlc_config.json
+++ b/mlc_config.json
@@ -1,0 +1,7 @@
+{
+  "ignorePatterns": [
+    {
+      "pattern": "^https://github.com/asyncapi/glee/blob/086c449/src"
+    }
+  ]
+}
--- a/mlc_config.json
+++ b/mlc_config.json
@@ -2,6 +2,9 @@
   "ignorePatterns": [
     {
       "pattern": "^https://github.com/asyncapi/glee/blob/086c449/src"
+    },
+    {
+      "pattern": "^https://asyncapi.org/glee/docs"
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


mlc_config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "ignorePatterns": [
+    {
+      "pattern": "^https://github.com/asyncapi/glee/blob/086c449/src"
+    },
+    {
+      "pattern": "^https://asyncapi.org/glee/docs"
```
fmvilas, line 7: I would follow the same pattern as the rest of the tools:

```suggestion
      "pattern": "^https://www.asyncapi.com/tools/glee"
```

However, I'm not sure if we should even have that yet. We can add it later.

fmvilas (changes requested):

mlc_config.json
KhudaDad414: where do you think https://github.com/asyncapi/glee/blob/master/examples/social-network/notifications-service/README.md?plain=1#L15 should point to for now? since the current link is dead. 🤔

fmvilas: Oh! I didn't notice this one. There are no docs yet so I'd point it to this repo:
```suggestion
      "pattern": "^https://www.github.com/asyncapi/glee"
```

fmvilas (approved):
LGTM 👍

PR #318, opened 2022-06-06
feat: automatically generate docs

### LINKED ISSUE #261: Automatically generate docs
#### Reason/Context

When describing the benefits of a spec-driven framework like Glee, I'm always mentioning the fact that we could have docs and code always in sync. However, we're not generating any docs, we're leaving this responsibility to the user, which kinda breaks the purpose 😅 

#### Description

We should provide a way to automatically generate docs. Here are the ways that come to my mind:

1. Start an HTTP static server and serve the docs there. For instance, `http://localhost:3000/docs` would show the docs. Fully configurable from the glee.config.js file.
2. Generate docs in a folder inside the project. Fully configurable from glee.config.js file, like the path to the folder and the Generator template to use. We can default to `docs` and the https://github.com/asyncapi/markdown-template/.
3. A Github Action (maybe Gitlab CI too?) that calls a webhook URL (or multiple ones?) to update external systems, like developer portals. E.g., a POST call to a URL containing the AsyncAPI definition. This should be a separate package though.

AceTheCreator:
**Description**
This PR provides a way to automatically generate docs

It generates docs in a folder inside the project. Fully configurable from the glee.config.js file, like the path to the folder and the Generator template to use. We can default to docs and the https://github.com/asyncapi/markdown-template/.

**Related issue(s)**
Resolves #261


AceTheCreator:
> Looks good to me, we are only missing a test to ensure it works slightly_smiling_face There is [a bunch of middleware test](https://github.com/asyncapi/glee/tree/master/test/middlewares) you can look for guidance, otherwise let me know if you get stuck or have any questions slightly_smiling_face

@jonaslagoni please kindly review this PR :)

test/middlewares/generateDocs.test.ts
```diff
@@ -42,18 +42,28 @@ const CONFIG_TEST_DATA = {
...
+        CONFIG_TEST_DATA,
+        testDir
+      );
+      expect(result).toBe('done');
+    } catch (e) {
+      err = e;
+    }
+
+    expect(err).toBeUndefined();
+    fs.emptyDirSync(testDir);
```
magicmatatjahu, line 67: Please at the end remove the `testDir` dir, not only remove content from it.

AceTheCreator: @magicmatatjahu  what do you mean by "not only remove content from it"?

magicmatatjahu: Read that https://github.com/jprichardson/node-fs-extra/blob/master/docs/emptyDir-sync.md. You should remove dir.

magicmatatjahu: Ok.. my fault. You shouldn't. Don't focus on my previous comments.

AceTheCreator: Okay 😂. Is the PR okay?

magicmatatjahu: I don't know if PR is okay, but that test is.

src/index.ts
```diff
@@ -41,7 +44,7 @@ export default async function GleeAppInitializer () {
     showAppDir: GLEE_PROJECT_DIR !== process.cwd(),
     showFunctionsDir: GLEE_FUNCTIONS_DIR !== resolve(GLEE_DIR, 'functions'),
   })
-
+  const gleeDocs = await readGleeConfig(GLEE_CONFIG_FILE_PATH)
```
fmvilas, line 47: I don't think you need the `readGleeConfig` function. You already have the config loaded a few lines before.

src/middlewares/generateDocs.ts
```diff
@@ -0,0 +1,36 @@
...
+  )
+  try {
+    await generator.generateFromString(JSON.stringify(resolvedData))
+    logLineWithIcon(
+      ':zap:',
+      'Successfully generated docs for your specification'
+    )
+    return 'done'
+  } catch (error) {
+    logLineWithIcon(
```
fmvilas, line 26: You can use this instead: https://github.com/asyncapi/glee/blob/master/src/lib/logger.ts#L123.

fmvilas (changes requested):
Left a couple of comments. I think this PR could be simpler. If I'm understanding correctly, you don't need the `readGleeConfig` util function. We already have the `initializeConfig` function that does the same and it's actually doing it with the config only.

src/middlewares/generateDocs.ts
AceTheCreator: @fmvilas I just updated the PR. Kindly review it :)

src/index.ts
```diff
@@ -120,6 +121,7 @@ export default async function GleeAppInitializer () {
   })
 
   app.on('adapter:server:ready', (e: EnrichedEvent) => {
+    generateDocs(parsedAsyncAPI, config, null)
```
fmvilas, line 124: Why here? If we generate docs on "adapter:server:ready", it will only work for WebSocket and HTTP servers but not for clients (MQTT, Kafka, etc.). I think it should be generated as soon as possible and not necessarily when an event happens.

src/middlewares/generateDocs.ts
```diff
@@ -0,0 +1,31 @@
...
+    path.resolve(
+      resDir ? resDir : './',
+      configData && configData.folder ? configData.folder : 'docs'
+    )
+  )
+  try {
+    await generator.generateFromString(JSON.stringify(resolvedData))
+    logLineWithIcon(
+      ':zap:',
+      'Successfully generated docs for your specification'
```
fmvilas, line 22: "specification" = https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md
"document" = https://github.com/asyncapi/spec/blob/master/examples/simple.yml

People don't have specs but AsyncAPI documents. I'd keep it simpler and rephrase it as:

```suggestion
      'Successfully generated docs'
```

src/middlewares/generateDocs.ts
```diff
@@ -0,0 +1,31 @@
...
+    )
+    return 'done'
+  } catch (error) {
+    logErrorLine(
+      `Failed to generate docs for your specs due to the following reason: ${error}`,
+    )
```
fmvilas, line 28: If you're going to log the reason, i.e., the full error, I recommend you use [logError](https://github.com/asyncapi/glee/blob/master/src/lib/logger.ts#L129). Have a look at the `logger.ts` file and get familiar with the stuff there 😄 

```suggestion
    logErrorLine('Failed to generate docs...')
    logError(error)
```

test/lib/util.test.ts
```diff
@@ -49,4 +51,27 @@ describe('util', () => {
       expect(functionEvent.glee).toBe(glee)
     })
   })
+  describe('readGleeConfig', () => {
```
fmvilas, line 54: 🚜 ...
😄

test/middlewares/generateDocs.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    folder: 'output',
+    template: 'markdown-template',
+  },
+}
+
+describe('generateDocs', () => {
+  it('should generate documentation', async () => {
+    const testDir = tmpdir() + `/${CONFIG_TEST_DATA.generator.folder}`;
+    fs.emptyDirSync(testDir);
+    let err: Error | undefined = undefined;
```
fmvilas, line 53: I think you can do something like this:

```suggestion
    let err?: Error
```

or like this, I can't remember:

```suggestion
    let err: Error?
```

test/middlewares/generateDocs.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    url: 'redis://localhost:6379',
+  },
+  generator: {
+    folder: 'output',
+    template: 'markdown-template',
+  },
+}
+
+describe('generateDocs', () => {
+  it('should generate documentation', async () => {
```
fmvilas, line 50: What is this test testing exactly? 🤔

fmvilas (changes requested):

test/middlewares/generateDocs.test.ts
AceTheCreator: @fmvilas I don't think I understand what you're trying to do here but it seems your code snippet is wrong :) 

Mind explaining what you're trying to achieve with that snippet?

test/middlewares/generateDocs.test.ts
fmvilas: LOL man, I know it's testing the `generateDocs` function but my question was more about the what or how is it exactly testing it. I have the feeling that it's not testing much.

test/middlewares/generateDocs.test.ts
fmvilas: I was trying to do something like this: https://www.typescriptlang.org/docs/handbook/advanced-types.html#optional-parameters-and-properties. But it looks like it only works on parameters and object properties 🤦

fmvilas: In any case, you can simplify it like this:

```suggestion
    let err: Error | undefined
```

test/middlewares/generateDocs.test.ts
fmvilas: Yeah, I have the feeling that this test is not "testing" much. I mean, it may generate something but that doesn't mean that what's generated is correct right. I know it's a unit test and not an integration test but this test feels a bit useless to me 😅

AceTheCreator: So, are you suggesting I get rid of the test?

test/lib/util.test.ts
AceTheCreator: Yes, you're right :)

test/middlewares/generateDocs.test.ts
fmvilas: Yeah but I'm a bit concerned that I'm probably not seeing the big picture here and this test should be kept, that's why I was asking. Don't have a strong opinion if it should stay or not so up to you 😄

fmvilas: Yeah, don't go that far. That would be an integration test and would be a pain to maintain for no big benefit. Leave it as is, @AceTheCreator, that's fine 👍

jonaslagoni: If you don't want to depend on the generator, you could mock `generator.generateFromString` to just return whatever you desire, and then check it created everything as expected.

So it's more tuned towards unit testing 🙂

AceTheCreator: @jonaslagoni I could just leave it as it is now :)

src/middlewares/generateDocs.ts
```diff
@@ -0,0 +1,26 @@
+import path from 'path'
+import { logInfoMessage, logLineWithIcon, logError } from '../lib/logger.js'
+import Generator from '@asyncapi/generator'
+
+export default async (spec, config, resDir) => {
+  logInfoMessage(`Generating docs for your parsed specification...`)
+  const configData = config.generator
+  const resolvedData = spec._json
```
fmvilas, line 8:
```suggestion
  const resolvedData = spec.json()
```

src/middlewares/generateDocs.ts
```diff
@@ -0,0 +1,26 @@
...
+      ? `@asyncapi/${configData.template}`
+      : '@asyncapi/markdown-template',
+    path.resolve(
+      resDir ? resDir : './',
+      configData && configData.folder ? configData.folder : 'docs'
+    )
+  )
+  try {
+    await generator.generateFromString(JSON.stringify(resolvedData))
+    logLineWithIcon(":zap:", "Successfully generated docs")
```
fmvilas, line 20: I know you're excited about it but a zap might be too much for generated docs 😄 Especially, because we're also using it for when the service connects to the broker.

```suggestion
    logInfoMessage('Successfully generated docs')
```

src/middlewares/generateDocs.ts
```diff
@@ -0,0 +1,26 @@
...
+    return 'done'
```
fmvilas, line 21: Why are you returning here? And why are you returning the string `'done'`? 🤔

test/middlewares/generateDocs.test.ts
```diff
@@ -0,0 +1,60 @@
...
+const CONFIG_TEST_DATA = {
+  websocket: {
+    httpServer: 'customServer',
+    adapter: 'native', // Default. Can also be 'socket.io' or a reference to a custom adapter.
+  },
+  cluster: {
+    adapter: 'redis',
+    name: 'gleeCluster',
+    url: 'redis://localhost:6379',
+  },
```
fmvilas, line 42: Can't you just provide the necessary data for your test? 🤔

test/middlewares/generateDocs.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  generator: {
+    folder: 'output',
+    template: 'markdown-template',
```
fmvilas, line 45: This should fail, doesn't it? The template is called `@asyncapi/markdown-template`.

src/middlewares/generateDocs.ts
```diff
@@ -0,0 +1,26 @@
...
+import Generator from '@asyncapi/generator'
+
+export default async (spec, config, resDir) => {
+  logInfoMessage(`Generating docs for your parsed specification...`)
+  const configData = config.generator
+  const resolvedData = spec._json
+  const generator = new Generator(
+    configData && configData.template
+      ? `@asyncapi/${configData.template}`
+      : '@asyncapi/markdown-template',
```
fmvilas, line 12: Oh! After checking tests I realized about this. Don't scope templates to `@asyncapi`. People might want to use a template of their own:

```suggestion
    configData?.template || '@asyncapi/markdown-template',
```

fmvilas (changes requested):
More feedback after summer 😅 Sorry for taking this long, I completely forgot this PR existed.

AceTheCreator:
Updated the PR, any other suggestions? @fmvilas

AceTheCreator:
> Left just one comment.
> 
> I'd also like to mention that, what would make this perfect would be to have a `glee docs` command that triggers the generation of the docs without having to run `glee dev` or `glee start`. So, if you're in the mood, just add it here: https://github.com/asyncapi/glee/blob/master/src/cli/index.ts. Then people just have to have a package.json script like `npm run docs` that will actually call `glee docs`, the same as we do with the other commands at https://github.com/asyncapi/glee/blob/master/examples/dummy/package.json#L6.

By extending it to meet this requirement, do you feel It still should still reside in the middleware dir?

AceTheCreator:
@fmvilas updated the PR. wyt?

src/cli/index.ts
```diff
@@ -16,27 +19,40 @@ if (command === 'dev') {
...
   import('./start.js')
+} else if (command === 'docs') {
+  (async () => {
+    const config = await initializeConfigs()
+    const parsedAsyncAPI = await getParsedAsyncAPI()
+    generateDocs(parsedAsyncAPI, config, null)
+  })()
 } else {
-  console.error(`Unknown command "${args[0]}"`)
+  console.error(`Unknown command '${args[0]}'`)
```
fmvilas, line 55: I don't think this change is right. The double quotes are more appropriate to convey "literal meaning" IMHO.

src/middlewares/generateDocs.ts
```diff
@@ -3,8 +3,9 @@ import { logInfoMessage, logError } from '../lib/logger.js'
 import Generator from '@asyncapi/generator'
 
 export default async (spec, config, resDir) => {
+  const configData = config.docs
+  if(configData && configData.generate){
```
fmvilas, line 7:
```suggestion
if (configData?.generate) {
```

src/cli/index.ts
```diff
@@ -16,27 +19,40 @@ if (command === 'dev') {
...
+      )
+    },
   })
 } else if (command === 'start') {
   import('./start.js')
+} else if (command === 'docs') {
+  (async () => {
+    const config = await initializeConfigs()
+    const parsedAsyncAPI = await getParsedAsyncAPI()
+    generateDocs(parsedAsyncAPI, config, null)
```
fmvilas, line 52: It would be awesome if we could just have this:

```suggestion
    generateDocs()
```

Probably another function in another file called `src/docs.ts`. This way we hide the complexity from this file.

test/middlewares/generateDocs.test.ts
```diff
@@ -31,15 +31,16 @@ const TEST_ASYNCAPI_DOCUMENT = new AsyncAPIDocument({
 })
 
 const CONFIG_TEST_DATA = {
-  generator: {
+  docs: {
+    generate: false,
```
fmvilas, line 35: I know I suggested `generate` but what do you think of `enabled` instead? 🤔

fmvilas (changes requested):

fmvilas (approved):
LGTM 👍

Maybe @KhudaDad414 and @Souvikns want to have a look as well.

src/lib/docs.ts
```diff
@@ -0,0 +1,27 @@
+import path from 'path'
+import { logInfoMessage, logError } from './logger.js'
+import Generator from '@asyncapi/generator'
+
+export default async (spec, config, resDir) => {
+  const configData = config.docs
```
KhudaDad414, line 6: @AceTheCreator I don't see this configuration documented anywhere. am I missing something?

KhudaDad414 (changes requested):

src/cli/index.ts
```diff
@@ -16,25 +17,34 @@ if (command === 'dev') {
...
+          },
+        }
+      )
+    },
   })
 } else if (command === 'start') {
   import('./start.js')
+} else if (command === 'docs') {
+  docs()
 } else {
```
Souvikns, line 48: I am unable to create docs, some documentation or example would be awesome. I am using `glee docs` and `glee start` but nothing works I think I need to add something in the `glee.config.js` which I don't find any documentation of.

Souvikns (changes requested):
Overall awesome, just missing documentation.

AceTheCreator:
Updated the config docs cc @KhudaDad414 @Souvikns

docs/config-file.md
```diff
@@ -31,6 +31,11 @@ Here is an example of a `glee.config.js` file for reference:
 export default async function () {
   return {
     websocket: {
+      docs: {
+        enabled: false // Enable/Disable documentation generation
+        folder: 'output' // Folder where you want the output of your docs to reside.
+        template: '@asyncapi/markdown-template' // Type of template you want to use.
+      }
```
Souvikns, line 38:
```suggestion
export default async function () {
  return {
        docs: {
          enabled: false // Enable/Disable documentation generation
          folder: 'output' // Folder where you want the output of your docs to reside.
          template: '@asyncapi/markdown-template' // Type of template you want to use.
      }
    websocket: {
```

It should be this right, I think docs should not go under any protocol-specific configuration, instead should be a global config. Also, I think we should by default create docs, and let user `enabled: false` it if required.

src/lib/docs.ts
Souvikns: @AceTheCreator you can update the type here https://github.com/asyncapi/glee/blob/0d4ff218ad2fec369fb77a9fcc6496424b292ef5/src/lib/index.d.ts#L36-L40

docs/config-file.md
AceTheCreator: > It should be this right, I think docs should not go under any protocol-specific configuration, instead should be a global config

Thanks for pointing that out, it was a mistake 😄

AceTheCreator: Updated the PR @Souvikns

Souvikns (dismissed):

src/lib/docs.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (configData?.enabled) {
```
fmvilas, line 7: With this check, docs will only be generated if we add this configuration to the config file. It would be awesome if, by default, it generated docs and then people can disable it if they want to. So maybe it should be `configData?.disabled` instead? Folder can default to `docs` and template to `@asyncapi/markdown-template`.

fmvilas (changes requested):

src/lib/docs.ts
AceTheCreator: Why should it generate if it says `configData?.disabled` ? isn't that saying generate if disabled?

AceTheCreator: shouldn't i be doing `!configData || configData?.enabled`?

src/lib/docs.ts
```diff
@@ -0,0 +1,27 @@
+import path from 'path'
+import { logInfoMessage, logError } from './logger.js'
+import Generator from '@asyncapi/generator'
+
+export default async (spec, config, resDir) => {
+  const configData = config.docs
+  if (!configData || configData?.enabled) {
```
AceTheCreator, line 7: Something like this @fmvilas

docs/config-file.md
```diff
@@ -30,6 +30,11 @@ Here is an example of a `glee.config.js` file for reference:
 ```js
   return {
+    docs: {
+      enabled: true // Enable/Disable documentation generation
+      folder: 'docs' // Folder where you want the output of your docs to reside.
+      template: '@asyncapi/markdown-template' // Type of template you want to use.
+    }
```
KhudaDad414, line 37:
```suggestion
      enabled: true, // Enable/Disable documentation generation
      folder: 'docs', // Folder where you want the output of your docs to reside.
      template: '@asyncapi/markdown-template' // Type of template you want to use.
    }
```

KhudaDad414 (changes requested):
LGTM! left some suggestions.

package.json
```diff
@@ -36,6 +36,8 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@asyncapi/generator": "^1.9.3",
```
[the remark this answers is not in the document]
AceTheCreator: Thanks for the review @KhudaDad414

KhudaDad414 (dismissed):
🚢 🙇

AceTheCreator:
@fmvilas @Souvikns any other suggestions?

src/lib/docs.ts
fmvilas: What about the case when `configData` exists but `enabled` is not defined? It should still work but with this check it won't. The reason I said to use `disabled` instead is that you can easily check for it like this:

```js
  if (configData?.disabled) return
```

This way, you early return **only** when you find that `disabled` is true.

fmvilas: Alternatively, you can check for this:

```js
  if (configData?.enabled === false) return
```

That should work as well, I think.

fmvilas: But seems more logical to me that, since the default is that it generates docs, you only go there when you want to disable it or when you want to change the template.

AceTheCreator: Ah! Now I understand your point @fmvilas

AceTheCreator: Updated the PR @fmvilas. wyt?

Souvikns (dismissed):

src/lib/docs.ts
fmvilas: @AceTheCreator I think you missed this conversation.

AceTheCreator:
Updated the ReadME @fmvilas. wyt?

fmvilas (approved):
LGTM 👍

fmvilas:
@jonaslagoni will have to approve too.

PR #319, opened 2022-06-21
feat: add ws client adapter 

### LINKED ISSUE #259: Add support for WebSocket clients

Souvikns:
**Description**
This PR is trying to implement a `websocket` client adapter for glee. 

So the main idea is, if I have a WebSocket server, my glee application should be able to - 
- connect to that server 
- receive and send data
- validate the data received or sent according to the specification file. 

**Related issue(s)**

Fixes #259


src/adapters/ws/client.ts
```diff
@@ -0,0 +1,78 @@
...
+            })
+
+
+            this.client.on('message', (data) => {
+                /**
+                 * For POC I used hard coded chanel name,
+                 * we need to dynamically figure this out. 
+                 */
+                const msg = this._createMessage('listen', data);
+                this.emit('message', msg, this.client);
```
Souvikns, line 54: @fmvilas I am kinda stuck here. 

Currently, we create adapters according to the number of servers we define in the specification file. This works well in the case of servers as we can parse the URL from the request body to figure out which channel to forward the message but in the case of the client, multiple channels could subscribe to different WebSocket servers or to different paths in the same server (something I am doing in my example). 

Can I somehow reference server in channel, I think this is something v3 will be bringing, but right now how do we determine the channel.

src/adapters/ws/client.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    async connect(): Promise<this> {
+        return this._connect();
+    }
+
+    async send(message: GleeMessage) {
+        return this._send(message);
+    }
+
+    _connect(): Promise<this> {
```
fmvilas, line 24: You may want to make this private as well. When I did the first adapters, JS didn't support private stuff on classes 😄

src/adapters/ws/client.ts
```diff
@@ -0,0 +1,78 @@
...
+        return new Promise((resolve) => {
+
+            const subscribedChannels = this.getSubscribedChannels();
```
fmvilas, line 27: This is not being used so it means you're not using the channels information at all. Check my comment below 👇

src/adapters/ws/client.ts
fmvilas: Replied in other lines to point you in the right direction.

src/registerAdapters.ts
```diff
@@ -35,7 +36,21 @@ function registerAdapterForServer(serverName: string, server: Server, app: Glee,
...
-    if (!configWsAdapter || configWsAdapter === 'native') {
+    /**
+     * We access the `x-kind` parameter from server object
+     */
+    const kind = server.json('x-kind');
+    /**
+     * If the server is remote we create client adapter otherwise 
+     * we create server adapter. 
+     */
+    if (kind === 'remote') {
```
fmvilas, line 47: I think we should default to client whenever possible. Glee is a client for multiple protocols. It should be an exception that it behaves as a server IMHO.

fmvilas (changes requested):

src/adapters/ws/client.ts
```diff
@@ -0,0 +1,78 @@
...
+            const securityRequirements = (this.AsyncAPIServer.security() || []).map(sec => {
+                const secName = Object.keys(sec.json())[0]
+                return this.parsedAsyncAPI.components().securityScheme(secName)
+            })
+
+            /**
+             * We do not spin up a server and just create a ws client 
+             * to connect to a existing ws server. 
+             */
+            const url = new URL(this.AsyncAPIServer.url() );
```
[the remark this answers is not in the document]
Souvikns: I thought of doing this, but since glee could be a server or a client at the same time -

we could have spec like this 

```yaml
asyncapi: 3.0.0
servers:
  client:
    url: ws: localhost:3001
    protocol: ws
  local: 
    url: ws:localhost:4000
    protocol: ws
    x-kind: local
channels:
  user/signedup: 
    ...
  testClient:
    ...
```

Now how do we figure out which server these channels belong to? or is it bad practice to have more than one server per `asyncapi` file, because I think if my glee application is able to connect to other ws servers then I can connect to multiple servers at once.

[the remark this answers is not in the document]
fmvilas: First of all, this example is easy, you should ignore the server that has `x-kind: local` because you don't want to create a client for a server that's "local". If it's local, then we should create a server, not a client.

That said, the AsyncAPI file above, means that `user/signedup` and `testClient` exist in all severs. If you want to restrict channels to certain servers, you can do it using the [`servers` attribute in the Channel Item Object](https://www.asyncapi.com/docs/reference/specification/v2.4.0#channelItemObject).

Souvikns:
@fmvilas need your review again

jonaslagoni:
@fmvilas as you started the review process, I am assuming you wish to finish it 🙂

.eslintignore
```diff
@@ -1,5 +1,6 @@
 node_modules
 examples/social-network/frontend
+examples/ws-server
```
fmvilas, line 3: If you end up needing the `ws-server` example, I recommend you move this line to a `.eslintignore` file inside `examples/ws-server`. Probably we should do the same with the line above.

Souvikns: I am removing the `ws-server` example as I don't need it anymore.

Souvikns:
One weird issue I am running into while sending headers from the client side, is the headers are converted to lowercase. 

```ts
{
  'sec-websocket-version': '13',
  'sec-websocket-key': 'nL+pNzTS+0ryHcEU/4ccPQ==',
  connection: 'Upgrade',
  upgrade: 'websocket',
  authorization: 'bearer arb-tokenValue',
  'sec-websocket-extensions': 'permessage-deflate; client_max_window_bits',
  host: 'localhost:3000'
}
```
Whereas I am sending the `Authorization` header and it is received as `authorization`
```yaml
    bindings:
      ws:
        bindingVersion: 0.1.0
        headerValues:
          Authorization: 'bearer $TOKEN'
```

Found this https://stackoverflow.com/questions/50477233/issue-with-netty-websocket-headers-name-in-lowercase

fmvilas (changes requested):

examples/crypto-websockets/client/asyncapi.yaml
```diff
@@ -0,0 +1,35 @@
...
+info: 
+  title: asyncapicoin client
+  version: 1.0.0
+  description: |
+    This app creates a client that subscribes to the server for the price change.
+servers:
+  websockets:
+    url: ws://localhost:3000
+    protocol: ws
+    x-kind: remote
```
fmvilas, line 11: Let's move this property out of the server definition because that's how it will be done in v3.0.0. For now, we can just use our own extension like this:

```suggestion
x-remoteServers:
  - websockets
```

This will change the logic in the `registerAdapters.ts` file.

examples/crypto-websockets/server/asyncapi.yaml
```diff
@@ -0,0 +1,38 @@
...
+info:
+  title: asyncapicoin server
+  version: 1.0.0
+  description: |
+    This app creates a dummy server that would stream the price of a fake crypto currency
+servers:
+  websocket:
+    url: ws://localhost:3000
+    protocol: ws
+    x-kind: local
```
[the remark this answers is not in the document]
fmvilas: Let's move this property out of the server definition because that's how it will be done in v3.0.0. For now, we can just use our own extension like this:

```suggestion
x-localServers:
  - websocket
```

This will change the logic in the `registerAdapters.ts` file.

fmvilas (changes requested):
Almost there. Just left a suggestion to make remote/local server stuff more similar to what it would be in v3.

src/registerAdapters.ts
```diff
@@ -26,6 +26,7 @@ export default async (app: Glee, parsedAsyncAPI: AsyncAPIDocument, config: GleeC
 
 function registerAdapterForServer(serverName: string, server: Server, app: Glee, parsedAsyncAPI: AsyncAPIDocument, config: GleeConfig) {
   const protocol = server.protocol()
+  const remoteServers = parsedAsyncAPI.extensions()['x-remoteServers']
```
fmvilas, line 29:
```suggestion
  const remoteServers = parsedAsyncAPI.extension('x-remoteServers')
```

src/adapters/ws/client.ts
```diff
@@ -33,50 +33,51 @@ class WsClientAdapter extends Adapter {
...
-        const { query, headers } = this.getBindingValues(
-          queryValues,
-          headerValues
-        )
+  private async _connect(): Promise<this> {
+    const channelsOnThisServer = this.getWsChannels()
+
+    for (const channel of channelsOnThisServer) {
+      const config = await this.resolveProtocolConfig('ws')
+      const clientConfig = config?.client;
```
fmvilas, line 41: Not used:

```suggestion
```

fmvilas (changes requested):

src/lib/index.d.ts
```diff
@@ -18,8 +18,9 @@ export type WebsocketAdapterConfig = {
     port?: number,
   },
   client?: {
+    query?: any
     authentication?: {
```
Souvikns, line 22:
```suggestion
   headers?: any
```

Instead of adding authentication should we just allow header and query config, with headers the user is able to set any type of authentication he desires?

fmvilas: But the user can already do that once you merge [this PR of yours](https://github.com/asyncapi/bindings/pull/144/files), right? We only saw the need for authentication because it usually contains secrets.

Souvikns: Yeah, just a suggestion, currently we have an `authentication` config and we are creating a header `Authentication: bearer TOKEN`, and it works but maybe we just allow headers instead of authentication, and then we have a broader control, but we only have the need for authentication. We can always change it in the future if there is a need.

fmvilas: We'd have broader control at Glee but we'd be breaking the purpose of the AsyncAPI specification. If the `headers` values are in the AsyncAPI document, other tools will be able to see them. If they're in the Glee config, only Glee will be able to see them. You see the problem?

For instance, in #261 we want to use Generator to generate docs automatically. If the info is not in the AsyncAPI file, it will not end up in the documentation. And the same will happen for all the tools that the user might want to apply to their AsyncAPI file.

fmvilas (approved):
LGTM 👍 🚀

Souvikns:
@fmvilas sentiment analysis is failing, can you rerun the tests, I can't merge right now.

fmvilas:
@Souvikns done. I hope your sentiments are fine now 😜

PR #320, opened 2022-06-26
feat: add http adapter support for glee

ritik307:
### Description
Implement the HTTP adapter for both server and client

### Aims to -
- connect HTTP server as per the specification.
- Handle incoming messages only on the channels specified in the specification
- Send appropriate responses from the registered function.
- connect the HTTP client as per the specification.
- HTTP client should be able to connect via any other protocol and provide response in the specified registered function.

### Steps to test the codeflow
**Steps to run the base code**
1. `npm i `
2. `npm run dev`

**Steps to run the server**
4. cd `examples/anime-http/server`
5. `npm i`
6. `npm run dev`

**Steps to run the client**
8. cd `examples/anime-http/client`
9. `npm i`
10. `npm run dev`


examples/simple/asyncapi.yaml
```diff
@@ -0,0 +1,39 @@
+asyncapi: 2.0.0
+info:
+  title: Account Service
+  version: 1.0.0
+  description: This service is in charge of the process
```
fmvilas, line 5: Maybe you want to update title and description? It's not related to account processing anymore.

examples/simple/.glee/functions/onTest.js
```diff
@@ -0,0 +1,9 @@
+export default function (event) {
```
fmvilas, line 1: We should add `.glee` folders to `.gitignore`.

examples/simple/asyncapi.yaml
```diff
@@ -0,0 +1,39 @@
+asyncapi: 2.0.0
```
fmvilas, line 1:
```suggestion
asyncapi: 2.4.0
```

src/adapters/http/index.ts
```diff
@@ -0,0 +1,109 @@
...
+    name(): string {
+        return 'HTTP server'
+    }
+
+    async connect(): Promise<this> {
+        return this._connect();
+    }
+
+    async send(message: GleeMessage): Promise<void> {
+        console.log('control reached send');
```
fmvilas, line 20:
```suggestion
```

src/adapters/http/index.ts
```diff
@@ -0,0 +1,109 @@
...
+        return this._send(message);
+    }
+
+    _connect(): Promise<this> {
+        return new Promise((resolve, reject) => {
+            const serverUrl = new URL(this.serverUrlExpanded);
+            const httpServer = this.glee.options?.websocket?.httpServer || http.createServer();
```
fmvilas, line 27: We shouldn't use the HTTP server from the WebSocket configuration. That's only for WebSocket servers. You should use something like this instead:

```suggestion
            const httpServer = this.glee.options?.http?.server || http.createServer();
```

src/adapters/http/index.ts
```diff
@@ -0,0 +1,109 @@
...
+            const asyncapiServerPort = serverUrl.port || 80;
+            const optionsPort = this.glee.options?.websocket?.port;
```
fmvilas, line 29:
```suggestion
            const optionsPort = this.glee.options?.http?.serverPort;
```

src/adapters/http/index.ts
```diff
@@ -0,0 +1,109 @@
...
+            const port = optionsPort || asyncapiServerPort;
+
+            httpServer.on('request', (req, res) => {
+                this.res = res;
```
fmvilas, line 33: Have you tested what happens when we receive 2 or more requests in a row? Doesn't `this.res` get overwritten?

src/adapters/http/index.ts
```diff
@@ -0,0 +1,109 @@
...
+                const {query: searchParams} = url.parse(req.url, true);
+                const httpChannelBinding = this.parsedAsyncAPI.channel(pathname).binding('http');
+
+                if (httpChannelBinding) {
+                    const { query, method } = httpChannelBinding;
+
+                    if(method) {
+                        if(req.method !== method) {
+                            this.emit('error', new Error(`${req.method} is not allowed for this url`));
+                            res.end('HTTP/1/1 400 Bad Request\r\n\r\n');
```
fmvilas, line 56:
```suggestion
                            res.end('HTTP/1.1 400 Bad Request\r\n\r\n');
```

src/adapters/http/index.ts
```diff
@@ -0,0 +1,109 @@
...
+                }
+
+                const {query: searchParams} = url.parse(req.url, true);
+                const httpChannelBinding = this.parsedAsyncAPI.channel(pathname).binding('http');
+
+                if (httpChannelBinding) {
+                    const { query, method } = httpChannelBinding;
+
+                    if(method) {
+                        if(req.method !== method) {
```
fmvilas, line 54:
```suggestion
                    if(method && req.method !== method) {
```

src/adapters/http/index.ts
```diff
@@ -0,0 +1,109 @@
...
+                            this.emit('error', new Error(`${req.method} is not allowed for this url`));
```
fmvilas, line 55: What about the typical error message:

```suggestion
                            this.emit('error', new Error(`Cannot ${req.method} ${pathname}`));
```

This should give the user something like `Cannot POST /test`.

src/adapters/http/index.ts
```diff
@@ -0,0 +1,109 @@
...
+
+            this.emit('server:ready', { name: this.name(), adapter: this });
+
+            resolve(this);
+        })
+    }
+
+    async _send(message: GleeMessage): Promise<void> {
+        this.res.write(message.payload);
+        this.res.end();
```
fmvilas, line 97: We should conditionally `end` the connection here. I mean, usually yeah, you close it but when you're doing Server-Sent Events (SSE), you don't want to close the connection but leave it open and let the client close it when it's done. It may need some extra work on the HTTP binding or checking the HTTP headers to see if it contains the necessary headers to do SSE.

src/adapters/http/index.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    _createMessage(eventName: string, payload: any) {
+        return new GleeMessage({
+            payload,
+            channel: eventName
```
fmvilas, line 103:
```suggestion
    _createMessage(pathName: string, payload: any) {
        return new GleeMessage({
            payload,
            channel: pathName
```

fmvilas (changes requested):
Left a few comments.

Along with those, I'm adding two more generic comments here:

1. What does "adding HTTP header init" means? 😄 
2. It would be great if you can consider making it clear when it should be creating an HTTP server and when it should be creating a client (client can be done in another PR). For instance, if the server is marked as "local", then it should create a server. Otherwise, it shouldn't. See an example here: https://github.com/asyncapi/glee/pull/319/files#diff-5a59976f2bbdbefa8fb1b10842622e9c61def3c47869b93b2d0d3d1374530734R21.

ritik307:
> Left a few comments.
> 
> Along with those, I'm adding two more generic comments here:
> 
> 1. What does "adding HTTP header init" means? smile
> 2. It would be great if you can consider making it clear when it should be creating an HTTP server and when it should be creating a client (client can be done in another PR). For instance, if the server is marked as "local", then it should create a server. Otherwise, it shouldn't. See an example here: https://github.com/asyncapi/glee/pull/319/files#diff-5a59976f2bbdbefa8fb1b10842622e9c61def3c47869b93b2d0d3d1374530734R21.

Sorry for that I will try to be more clear on the commit messages 🙂

ritik307:
@fmvilas Can you kindly provide me with some resources to study web sockets, I am baffled with web sockets at the moment.😟

fmvilas:
Sure, MDN has a great resource: https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API.

src/adapters/http/index.ts
ritik307: so should i check for headers before closing the connection?

examples/simple/asyncapi.yaml
```diff
@@ -0,0 +1,39 @@
...
+    bindings:
+      http:
+        bindingVersion: 0.1.0
+        type: request
+        method: GET
+        query:
+          type: object
+          properties:
+            name:
+              type: string
```
Souvikns, line 21: HTTP should support multiple methods for which I did start an initial work https://github.com/asyncapi/bindings/pull/131 I would suggest you look into it and move it forward.

src/adapters/http/index.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    async send(message: GleeMessage): Promise<void> {
+        return this._send(message);
+    }
+
+    _connect(): Promise<this> {
+        return new Promise((resolve, reject) => {
+            const serverUrl = new URL(this.serverUrlExpanded);
+            const httpServer = this.glee.options?.http?.server || http.createServer();
```
Souvikns, line 26: This would change in the future, as per https://github.com/asyncapi/glee/pull/351 so do give it a look and also comment if you need any specific config for HTTP adapter be it authentication.

Souvikns (changes requested):
Since you are adding a new adapter to support the HTTP server you have to update the docs accordingly.

fmvilas (changes requested):

ritik307:
@Souvikns I have changed the `examples/simple/asyncapi.yaml` file according to the docs that you have referred but i am not sure whether i used `$ref: '#/components/messages/testReply'` at correct place or not. Kindly check it and let me know if any further changes are required

examples/simple/asyncapi.yaml
```diff
@@ -0,0 +1,50 @@
+asyncapi: 2.4.0
```
Souvikns, line 1: I told you to update the example, create an actual proper example that would help the users understand how to use glee with http protocol. You can either delete the `simple` example and start new or update it. It is up to you.

examples/simple/asyncapi.yaml
```diff
@@ -0,0 +1,50 @@
...
+  description: Test payload schema
+servers:
+  http:
+    url: http://localhost:8080
+    protocol: http
+channels:
+  test:
+    subscribe:
+      message:
+        $ref: '#/components/messages/testReply'
```
Souvikns, line 14: Yeah it is in the right place 👍🏼

src/adapters/http/index.ts
Souvikns: @ritik307 As I have told you earlier, this has changed, now there is a function that will resolve and give you the required config object. 
You need to do something like this 


```suggestion
            const httpServer = await this.resolveProtocolConfig('ws')
```

src/adapters/ws/server.ts
```diff
@@ -43,7 +43,7 @@ class WebSocketsAdapter extends Adapter {
 
       if (!pathname.startsWith(serverUrl.pathname) && !pathname.startsWith(`/${serverUrl.pathname}`)) {
         socket.end('HTTP/1.1 404 Not Found\r\n\r\n')
-        const err = new Error(`A client attempted to connect to channel ${pathname} but this channel is not defined in your AsyncAPI file.`)
+        const err = new Error(`A client attempted to connect to channel ${pathname} but this channel is not defined in your AsyncAPI file. hello2`)
```
Souvikns, line 46: I don't think you meant to do these changes

examples/anime-http/server/lifecycle/updateAnimeList.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable no-undef */
```
Souvikns, line 1: I don't think you need a lifecycle event in your server example

src/adapters/http/server.ts
```diff
@@ -0,0 +1,122 @@
+import Adapter from "../../lib/adapter.js";
+import GleeMessage from "../../lib/message.js";
+import http from "http";
+import { validateData } from "../../lib/util.js";
+import GleeError from "../../errors/glee-error.js";
+import * as url from "url";
+
+class HttpAdapter extends Adapter {
+  private res: any[] = [];
```
Souvikns, line 9: This is very basic, you should this to cache connections, store the connection so that you can later distinguish when there are more than one HTTP server in the spec.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,29 @@
+import Adapter from "../../lib/adapter.js";
+
+const axios = require('axios');
+
+class HttpClientAdapter extends Adapter {
+
```
Souvikns, line 6: You need to cache the client connection, because there could be more than one client servers in spec

Souvikns (changes requested):
Left some comments, you are going in the right direction,

src/adapters/http/server.ts
```diff
@@ -0,0 +1,128 @@
...
+            }
+            if (query) {
+              const { isValid, humanReadableError, errors } = validateData(
+                searchParams.query,
+                query
+              );
+              if (!isValid) {
+                const err = new GleeError({ humanReadableError, errors });
+                this.emit("error", err);
+                res.end("HTTP/1.1 400 Bad Request\r\n\r\n");
```
Souvikns, line 79: I think instead of sending only `Bad Request` we should also send a formated version of the `humanReadableError` object as it will help the user fix the issue faster.

src/adapters/http/server.ts
```diff
@@ -0,0 +1,128 @@
...
+            }
+            if (body) {
+              const { isValid, humanReadableError, errors } = validateData(
+                searchParams.body,
+                body
+              );
+              if (!isValid) {
+                const err = new GleeError({ humanReadableError, errors });
+                this.emit("error", err);
+                res.end("HTTP/1.1 400 Bad Request\r\n\r\n");
```
Souvikns, line 91: Same here, we should send a formated version of the `humanReadableError` object as it will help the user fix the issue faster.

src/lib/adapter.ts
```diff
@@ -49,6 +49,7 @@ class GleeAdapter extends EventEmitter {
 
     this.on('error', err => { this._glee.injectError(err) })
     this.on('message', (message, connection) => {
+      console.log("--error message1--");
```
Souvikns, line 52: You should remove the console logs you were using to test.

src/lib/index.d.ts
```diff
@@ -25,6 +26,16 @@ export type WebsocketAdapterConfig = {
   }
 }
 
+export type HttpAdapterConfig = {
+  server: {
+    httpServer?: SyntaxError,
```
Souvikns, line 31:
```suggestion
    httpServer?: any,
```
why is it of type SyntaxError? Also you should update the [docs](https://github.com/asyncapi/glee/blob/master/docs/config-file.md)

Souvikns (changes requested):
Good progress so far, I left some comments. Remove all the log statements you were using for testing purposes. and update the [config docs](https://github.com/asyncapi/glee/blob/master/docs/config-file.md).

src/lib/index.d.ts
```diff
@@ -25,6 +26,16 @@ export type WebsocketAdapterConfig = {
...
+export type HttpAdapterConfig = {
+  server: {
+    httpServer?: any,
+    adapter?: HttpServerType | typeof GleeAdapter
+    port?: number
+  },
+  client?: {
+    endpoint: any
+  }
+}
```
Souvikns, line 38: I don't think this and the documentation match, can you check it once, in `docs`you have `query` and body by here you have `endpoint`.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,52 @@
...
+        },
+        query:{
+          name: "trendingAnime",
+          rating: "5",
+          studio: "teststudio",
+          genre: "testgenre",
+        }
+      };
+      this.emit("send", requestObj);
+    });
```
Souvikns, line 28: Is it being called from functions, I can see that it is being called when the client first starts but does it work when you return a `send` operation from the functions?

Souvikns (changes requested):
Left some comments, you have some code smells that you need to fix, also check lint

src/adapters/http/client.ts
```diff
@@ -0,0 +1,51 @@
...
+          rating: "5",
+          studio: "teststudio",
+          genre: "testgenre",
+        }
+      }
+      this._send(requestObj);
+      return this
+    // });
+  }
+  async _send(message) {
```
Souvikns, line 29: Try to make the Axios call in the _send function, use the server name and the channel name from he gleeMessage that you receive as a parameter in the `_send` function.

src/registerAdapters.ts
```diff
@@ -74,9 +76,25 @@ function registerAdapterForServer(serverName: string, server: Server, app: Glee,
...
+        server,
+        parsedAsyncAPI
+      })
+    }
+  }
+
+  else {
     // TODO: Improve error message with link to repo encouraging the developer to contribute.
-    throw new Error(`Protocol "${server.protocol()}" is not supported yet.`)
+    throw new Error(`Proxptocol "${server.protocol()}" is not supported yet.`)
```
Souvikns, line 97: I think this is a typo

Souvikns, line 97: Fix this typo

Souvikns (changes requested):
I left some comments about the HTTP client adapter, give it a try.

Souvikns (changes requested):
Looking good so far, left some minor comments.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,40 @@
...
+    const clientConfig = config?.client
+    headers["Authenticaton"] = clientConfig?.authentication?.token
+    return this
+  }
+  async _send(message: GleeMessage): Promise<void> {
+    const method = message.payload.method //get post put
+    const url = `http://localhost:${message.payload.port}/${message.serverName}`
+    const body = message.payload.body
+    const query = message.payload.query
+
```
Souvikns, line 26: you don't need method from the message, you should get that from the spec file, as you have defined it there.

Souvikns (changes requested):

ritik307:
@Souvikns I have implemented the changes that you have requested on client http adaptor but i am currently facing issue in the implementation of `this.emit("message")` to send response back to client.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,40 @@
...
+  async send(message: GleeMessage): Promise<void> {
+    return this._send(message)
+  }
+
+  async _connect(): Promise<this> {
+    const headers = {}
+    const config: HttpAdapterConfig = await this.resolveProtocolConfig("http")
+    const clientConfig = config?.client
+    headers["Authenticaton"] = clientConfig?.authentication?.token
```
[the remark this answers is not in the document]
ritik307: i am not sure about the above implementation if it is correct or not?

ritik307:
@Souvikns resolved merge conflicts

src/registerAdapters.ts
```diff
@@ -6,7 +6,9 @@ import SocketIOAdapter from './adapters/socket.io/index.js'
 import RedisClusterAdapter from './adapters/cluster/redis/index.js'
 import { getSelectedServerNames } from './lib/servers.js'
 import Glee from './lib/glee.js'
-import { GleeConfig, GleeClusterAdapterConfig } from './lib/index.js'
+import { GleeConfig, GleeClusterAdapterConfig } from './lib/index.d'
```
Souvikns, line 9:
```suggestion
import { GleeConfig, GleeClusterAdapterConfig } from './lib/index.js'
```

Souvikns (changes requested):
Tests are failing, you need to make a small change

ritik307:
@Souvikns Made the changes you have suggested. Please check

docs/config-file.md
```diff
@@ -69,6 +70,21 @@ export default async function () {
...
+        adapter: "native", 
+        port: process.env.PORT,
+      },
+      client: {
+        query: {
+          foo: 'bar'
+        },
+        body: {
+          foo: 'bar'
+        }
```
Souvikns, line 86: You are missing the authentication examples here.

docs/config-file.md
```diff
@@ -129,13 +145,25 @@ These configurations apply to Glee itself, rather than any specific protocol.
...
-|kafka.auth.rejectUnauthorized | Boolean flag for accepting the valid SSL certificates
-|kafka.auth.username| The username to use during authentication.
-|kafka.auth.password| The password to use during authentication.
+|kafka.authentication| Kafka authentication configuration|
+|kafka.authentication.key | Kafka Broker Key
+|kafka.authentication.cert| Client certificate
+|kafka.authentication.clientId| Kafka client Id for authentication
+|kafka.authentication.rejectUnauthorized | Boolean flag for accepting the valid SSL certificates
+|kafka.authentication.username| The username to use during authentication.
+|kafka.authentication.password| The password to use during authentication.
```
Souvikns, line 154:
```suggestion
|kafka.auth| Kafka authentication configuration|
|kafka.auth.key | Kafka Broker Key
|kafka.auth.cert| Client certificate
|kafka.auth.clientId| Kafka client Id for authentication
|kafka.auth.rejectUnauthorized | Boolean flag for accepting the valid SSL certificates
|kafka.auth.username| The username to use during authentication.
|kafka.auth.password| The password to use during authentication.
```

docs/config-file.md
```diff
@@ -129,13 +145,25 @@ These configurations apply to Glee itself, rather than any specific protocol.
...
+
+### Http Server
+
+|Field|Description|
+|--|--|
+|http.server|Http server-specific configurations|
+|http.client|Http client-specific configurations|
+|http.server.adapter| The Glee adapter to use for the Http server. Defaults to a "native" Http implementation.|
+|websocket.server.port| The port to use when binding the Http server. This is useful when your server is behind a proxy and the port exposed for consumption is not the same as the port your application should be bound to. Defaults to the port specified in the selected AsyncAPI server.|
+|http.client.authetication| Authentication variables for client|
```
Souvikns, line 164:
```suggestion
|http.client.auth| Authentication variables for client|
```

it would be best if you defined the auth object for HTTP, like token-based or basic authentication

package.json
```diff
@@ -43,6 +43,7 @@
     "@types/qs": "^6.9.7",
     "ajv": "^6.12.6",
     "async": "^3.2.0",
+    "axios": "^1.2.5",
```
Souvikns, line 46: I think you should use [got](https://www.npmjs.com/package/got) library for making API calls, as it is already being used in a different PR.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,72 @@
...
+      connection: http,
+      channel: this.channelNames,
+    })
+    return this
+  }
+  async _send(message: GleeMessage): Promise<void> {
+    const headers = {}
+    const config: HttpAdapterConfig = await this.resolveProtocolConfig("http")
+    const clientConfig = config?.client
+    headers["Authenticaton"] = clientConfig?.authentication?.token
```
Souvikns, line 32:
```suggestion
    headers["Authentication"] = clientConfig?.authentication?.token
```

src/lib/index.d.ts
```diff
@@ -46,6 +47,20 @@ export type WebsocketAdapterConfig = {
...
 }
 
+export type HttpAdapterConfig = {
+  server: {
+    httpServer?: any,
+    adapter?: HttpServerType | typeof GleeAdapter
+    port?: number
+  },
+  client?: {
+    authentication?: {
```
Souvikns, line 57:
```suggestion
    auth?: {
```

src/lib/index.d.ts
```diff
@@ -46,6 +47,20 @@ export type WebsocketAdapterConfig = {
...
+      token?: string
```
Souvikns, line 58: you auth object does not match the docs example.

Souvikns (changes requested):
Looking good 👍🏼, just some minor changes

ritik307:
@Souvikns Made the changes. Kindly check

src/lib/index.d.ts
```diff
@@ -46,6 +47,20 @@ export type WebsocketAdapterConfig = {
...
 }
 
+export type HttpAdapterConfig = {
+  server: {
+    httpServer?: any,
+    adapter?: HttpServerType | typeof GleeAdapter
+    port?: number
+  },
+  client?: {
+    auth?: {
```
Souvikns, line 57: auth could be a function as well, I would suggest you take inspiration other adapter configs.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,72 @@
...
+      adapter: this,
+      connection: http,
+      channel: this.channelNames,
+    })
+    return this
+  }
+  async _send(message: GleeMessage): Promise<void> {
+    const headers = {}
+    const config: HttpAdapterConfig = await this.resolveProtocolConfig("http")
+    const clientConfig = config?.client
```
Souvikns, line 31: This is not the optimal way to access the auth config now, check out https://github.com/asyncapi/glee/blob/8afea22cb3096b15f72f091670e27f48b79ab66a/src/adapters/ws/client.ts#L34

src/lib/index.d.ts
```diff
@@ -46,6 +47,20 @@ export type WebsocketAdapterConfig = {
   }
 }
 
+export type HttpAdapterConfig = {
+  server: {
+    httpServer?: any,
+    adapter?: HttpServerType | typeof GleeAdapter
```
Souvikns, line 53: I don't think you need this for HTTP

Souvikns (changes requested):
auth config is not implemented correctly, after https://github.com/asyncapi/glee/pull/374 auth config acts differently from before.

src/lib/index.d.ts
```diff
@@ -5,6 +5,7 @@ import GleeConnection from './connection.js'
 import Glee from './glee.js'
 
 type WebSocketServerType = 'native' | 'socket.io'
+type HttpServerType = 'native' | 'socket.io'
```
KhudaDad414, line 8: Can we have `socket.io` HTTP servers? I think these options need to be specified in the documentation. 🤔

Souvikns: I think we only have a native HTTP adapter, better to remove the adapter parameter from the HTTP config object.

Souvikns (approved):
Looks good 👍🏼

Souvikns:
You have to wait for @fmvilas's review.

examples/anime-http/client/functions/index.ts
```diff
@@ -0,0 +1,24 @@
+export default async function () {
+  const payload = {
+    body: {
+      name: "tets",
```
fmvilas, line 4:
```suggestion
      name: "test",
```

examples/anime-http/client/functions/index.ts
```diff
@@ -0,0 +1,24 @@
...
+      rating: 5,
+      studio: "teststudio",
+      genre: "testgenre",
+    },
+    query: {
+      name: "tets",
```
fmvilas, line 10:
```suggestion
      name: "test",
```

examples/anime-http/client/functions/index.ts
```diff
@@ -0,0 +1,24 @@
+export default async function () {
```
fmvilas, line 1: Let's use single quotes everywhere in this file. Just for consistency.

examples/anime-http/client/functions/trendingAnimeListRecieverController.ts
```diff
@@ -0,0 +1,3 @@
+export default async function (event) {
+  console.log("trending Anime: ",event.payload)
```
fmvilas, line 2:
```suggestion
  console.log('Trending Anime:', event.payload)
```

examples/anime-http/client/asyncapi.yaml
```diff
@@ -0,0 +1,93 @@
+asyncapi: 2.4.0
+info:
+  title: asyncapiImdb client
```
fmvilas, line 3:
```suggestion
  title: AsyncAPI IMDB client
```

examples/anime-http/client/asyncapi.yaml
```diff
@@ -0,0 +1,93 @@
...
+              - rating
+              - genre
+              - studio
+            properties:
+              name:
+                type: string
+                description: name of the anime.
+              rating:
+                type: string
+                description: rating of the show.
```
fmvilas, line 67:
```suggestion
                description: Rating of the show.
```

examples/anime-http/client/asyncapi.yaml
```diff
@@ -0,0 +1,93 @@
...
+            type: object
+            required:
+              - name
+              - rating
+              - genre
+              - studio
+            properties:
+              name:
+                type: string
+                description: name of the anime.
```
fmvilas, line 84:
```suggestion
                description: Name of the anime.
```

examples/anime-http/client/asyncapi.yaml
```diff
@@ -0,0 +1,93 @@
...
+              rating:
+                type: number
+                description: rating of the show.
```
fmvilas, line 87:
```suggestion
                description: Rating of the show.
```

examples/anime-http/server/asyncapi.yaml
```diff
@@ -0,0 +1,114 @@
+asyncapi: 2.4.0
+info:
+  title: asyncapiImdb server
```
fmvilas, line 3:
```suggestion
  title: AsyncAPI IMDB server
```

examples/anime-http/server/asyncapi.yaml
```diff
@@ -0,0 +1,114 @@
...
+              description: name of the anime.
+            rating:
+              type: number
+              description: rating of the show.
+            genre:
+              type: string
+              description: The genre of anime.
+            studio:
+              type: string
+              description: The studio of anime.
```
fmvilas, line 56: This is repeated in the client example. Please choose one of them and use `$ref` in the other one.

examples/anime-http/server/asyncapi.yaml
```diff
@@ -0,0 +1,114 @@
...
+                description: name of the anime.
+              rating:
+                type: number
+                description: rating of the show.
+              genre:
+                type: string
+                description: The genre of anime.
+              studio:
+                type: string
+                description: The studio of anime.
```
fmvilas, line 114: This is almost the same as above. Let's use `components/schemas` to avoid repetition.

examples/anime-http/server/functions/trendingAnimeController.ts
```diff
@@ -0,0 +1,17 @@
+export default async function (event) {
```
fmvilas, line 1: Let's use single quotes in this file for consistency.

package.json
```diff
@@ -53,6 +53,7 @@
     "emojis": "^1.0.10",
     "eslint-plugin-github": "^4.3.5",
     "eslint-plugin-security": "^1.4.0",
+    "got": "^12.5.3",
```
fmvilas, line 56: Why is this change here?

src/adapters/http/client.ts
```diff
@@ -0,0 +1,72 @@
+import Adapter from "../../lib/adapter.js"
```
fmvilas, line 1: Let's use single quotes in this file for consistency.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+  async connect(): Promise<this> {
+    return this._connect()
+  }
+
+  async send(message: GleeMessage): Promise<void> {
+    return this._send(message)
+  }
+
+  async _connect(): Promise<this> {
```
fmvilas, line 19: There's no need to have this function here. I'd move the whole logic to the `connect` function instead.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,72 @@
...
+    this.emit("connect", {
+      name: this.name(),
+      adapter: this,
+      connection: http,
+      channel: this.channelNames,
+    })
+    return this
+  }
+  async _send(message: GleeMessage): Promise<void> {
```
fmvilas, line 28: There's no need to have this function here. I'd move the whole logic to the `send` function instead.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,72 @@
...
+    const headers = {}
+    const config: HttpAdapterConfig = await this.resolveProtocolConfig("http")
+    const auth: HttpAuthConfig = await this.getAuthConfig(config.client.auth)
+    headers["Authentication"] = auth?.token
+    const serverUrl = this.serverUrlExpanded
+    this.channelNames.forEach(async (channelName) => {
+      const channelInfo = this.parsedAsyncAPI.channel(channelName)
+      const httpChannelBinding = channelInfo.binding("http")
+      const servers = channelInfo.servers()
```
fmvilas, line 37: `servers` is a misleading term here. I'd call it `channelServers` instead.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,72 @@
...
+      if (
+        httpChannelBinding &&
+        (!servers.length || servers.includes(message.serverName))
```
fmvilas, line 40: Please extract this line to a variable above. This is really hard to reason about.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,72 @@
...
+      ) {
+        const method = httpChannelBinding.method //get post put
```
fmvilas, line 42:
```suggestion
        const method = httpChannelBinding.method
```

There are many more methods that GET, POST, and PUT.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,72 @@
...
+        const url = serverUrl + "/" + channelName
```
fmvilas, line 43: Please use template literals instead.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,72 @@
...
+        const body = message.payload.body
+        const query = message.payload.query
```
fmvilas, line 45: You can use destructuring here:

```suggestion
        const { body, query } = message.payload
```

Also, please, assign the right type to these values.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,72 @@
...
+            const msg = this._createMessage(channelName, res.body)
+            this.emit("message", msg, http)
+          })
+          .catch((err) => {
+            this.emit("error", err)
+          })
+      }
+    })
+  }
+  _createMessage(channelName: string, payload: any) {
```
fmvilas, line 63: Let's mark this method as `private` and remove the `_` in the beginning.

src/adapters/http/client.ts
fmvilas: Also, you're getting the query parameters from the message payload? Why?

src/adapters/http/server.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  async connect(): Promise<this> {
+    return this._connect()
+  }
+
+  async send(message: GleeMessage): Promise<void> {
+    return this._send(message)
+  }
+
+  async _connect(): Promise<this> { // NOSONAR
```
fmvilas, line 23: Please, move this logic to the `connect` method instead. Also, why `NOSONAR` here?

src/adapters/http/server.ts
```diff
@@ -0,0 +1,132 @@
+import Adapter from "../../lib/adapter.js"
+import GleeMessage from "../../lib/message.js"
+import http from "http"
+import { validateData } from "../../lib/util.js"
+import GleeError from "../../errors/glee-error.js"
+import * as url from "url"
+
+class HttpAdapter extends Adapter {
+  private res = new Map()
```
fmvilas, line 9: Can we come up with a more meaningful name for `res`, please?

src/adapters/http/server.ts
```diff
@@ -0,0 +1,132 @@
...
+            `A client attempted to connect to channel ${pathname} but this channel is not defined in your AsyncAPI file. here`
+          )
+          this.emit("error", err)
+          return err
+        }
+        const { query: query } = url.parse(req.url, true)
+        const searchParams = {
+          query,
+          body,
+        }
```
fmvilas, line 57: Why is `body` a part of `searchParams`?

src/registerAdapters.ts
```diff
@@ -74,7 +76,23 @@ function registerAdapterForServer(serverName: string, server: Server, app: Glee,
       }
 
     }
-  } else {
+  }else if (['http', 'https'].includes(protocol)) {
+    if(remoteServers){
```
fmvilas, line 80: There may be remote servers but none of them using the HTTP protocol. You have to check for the protocol here too.

fmvilas (changes requested):

fmvilas:
@KhudaDad414 have a look too please.

package.json
ritik307: got is been used in other PRs of Glee so I have also used the same here to make API calls as suggested by @Souvikns

src/adapters/http/client.ts
ritik307: there is no support for query params in the GleeMessage constructor, so i passed both the fields under payload only as message

fmvilas: Let's explore where we put it, maybe in the GleeMessage object or maybe in the GleeConnection object. Definitely not in the payload as it's conditioning how we send and receive payloads. Where do you think it fits best?

ritik307: IMO `GleeMessage` should be the right place to add query as it already has other fields required for the message, so I have implemented a solution with that approach. Please have a look.

KhudaDad414 (dismissed):
LGTM!

src/lib/message.ts
```diff
@@ -30,7 +32,8 @@ class GleeMessage extends EventEmitter {
   private _outbound: boolean
   private _cluster: boolean
   private _params: { [key: string]: string }
-  
+  private _query: any
```
fmvilas, line 35: Any? I think it's pretty safe to assume that the type would be something like `{ [key: string]: string } | { [key: string]: string[] }`. All the query parameters are going to be received as strings independently of their initial type.

Let's try to make a bigger effort here instead of just throwing an `any` type, please.

src/lib/message.ts
```diff
@@ -42,6 +45,7 @@ class GleeMessage extends EventEmitter {
    * @param {GleeConnection} [options.connection] The connection through which the message will be sent or has been received.
    * @param {Boolean} [options.broadcast=false] Whether the message should be broadcasted or not.
    * @param {Boolean} [options.cluster=false] Whether the message is from a cluster adapter or not.
+   * @param {Any} [options.query] Message query.
```
fmvilas, line 48: Please elaborate a bit more on this description. Maybe mention this is just for HTTP query parameters, etc. This doesn't say much right now.

fmvilas (changes requested):
Left some comments

ritik307:
@fmvilas Made the changes you have suggested. Please check

fmvilas (changes requested):
Left some more comments. Please don't rush it out. Take your time to evaluate what's the best solution. For instance, I see you're repeating `{ [key: string]: string } | { [key: string]: string[] }` in multiple places. Maybe it's time to define it as a custom type to reuse it.

src/adapters/http/client.ts
```diff
@@ -0,0 +1,63 @@
...
+      const httpChannelBinding = channelInfo.binding('http')
+      const channelServers = channelInfo.servers()
+      const isChannelServers = !channelServers.length || channelServers.includes(message.serverName)
+      if (
+        httpChannelBinding && isChannelServers
+      ) {
+        const method = httpChannelBinding.method
+        const url = `${serverUrl}/${channelName}`
+        const body:any = message.payload
+        const query: any = message.query
```
fmvilas, line 38: We now know the type of `query` so I don't think it should be `any`.

src/adapters/http/server.ts
```diff
@@ -0,0 +1,131 @@
...
+        pathname = pathname.startsWith("/") ? pathname.substring(1) : pathname
+        if (!this.parsedAsyncAPI.channel(pathname)) {
+          res.end("HTTP/1.1 404 Not Found1\r\n\r\n")
+          const err = new Error(
+            `A client attempted to connect to channel ${pathname} but this channel is not defined in your AsyncAPI file. here`
+          )
+          this.emit("error", err)
+          return err
+        }
+        const { query: query } = url.parse(req.url, true)
```
fmvilas, line 52: Wouldn't it be enough like this?

```suggestion
        const { query } = url.parse(req.url, true)
```

src/adapters/http/server.ts
```diff
@@ -0,0 +1,131 @@
...
+        const msg = this._createMessage(pathname, payload, searchParams)
+        this.emit("message", msg, http)
+      })
+    })
+
+    httpServer.listen(port)
+    this.emit("server:ready", { name: this.name(), adapter: this })
+    return this
+  }
+  _checkHttpBinding(req, res, pathname, httpChannelBinding, searchParams, payload) {
```
fmvilas, line 83: I think you can use types here too. This is all `any` right now.

src/lib/functions.ts
```diff
@@ -26,6 +26,7 @@ const OutboundMessageSchema = {
     },
     channel: { type: 'string' },
     server: { type: 'string' },
+    query: {}
   }
```
fmvilas, line 30: Let's validate `query` here too. We now know its type.

src/lib/index.d.ts
```diff
@@ -46,6 +52,17 @@ export type WebsocketAdapterConfig = {
...
 }
 
+export type HttpAdapterConfig = {
+  server: {
+    httpServer?: any,
+    port?: number
+  },
+  client?: {
+    auth?:  HttpAuthConfig | AuthFunction<HttpAuthConfig>,
+    query?: any,
```
fmvilas, line 62: `query` shouldn't be `any`

src/lib/message.ts
```diff
@@ -123,17 +130,27 @@ class GleeMessage extends EventEmitter {
...
+    this._query = value
+  }
   /**
    * Sends the message back to the server/broker.
    *
    * @param {Object} options
    * @param {Any} [options.payload] The new message payload. Pass falsy value if you don't want to change it.
    * @param {Object|null} [options.headers] The new message headers. Pass null if you want to remove them.
    * @param {String} [options.channel] The channel where the reply should go to.
+   * @param {Object} [options.query] The new message query. Pass falsy value if you don't want to change it.
```
fmvilas, line 147:
```suggestion
   * @param {Object} [options.query] The new message query parameters. Pass a falsy value if you don't want to change them.
```

fmvilas (changes requested):
Left a few more suggestions

src/lib/functions.ts
ritik307: I have created query as `object` type . And since we dont know what the user is going to pass in the query param i cant define the type of keys and values here. I have added the implementation according to my understanding please have a look.

fmvilas (approved):
LGTM 👍

Souvikns (approved):

PR #321, opened 2022-07-09
fix: ensure backward compatibilty while processing errors

### LINKED ISSUE #268: Glee fails when received or sent data doesn't match definition
#### Describe the bug
When we receive or try to send a message that doesn't match the definition in the AsyncAPI file, Glee breaks without showing the error and instead shows an `UnhandledPromiseRejectionWarning: TypeError` error.

#### How to Reproduce
Go to `social-network` example and change `postId` to be a string instead of a number. Either do it in the code or in the AsyncAPI file but not both. We want the code to do something different than what AsyncAPI file says, just so it fails.

This error appears:
Captura de pantalla 2022-03-14 a las 7 46 26

#### Expected behavior
It should show a nice error saying that the message we're receiving or trying to send is malformed. In essence, the [errorLogger middleware](https://github.com/asyncapi/glee/blob/master/src/middlewares/errorLogger.ts) should come into play.

#### Hints

It looks like [the variable `emws`](https://github.com/asyncapi/glee/blob/master/src/lib/glee.ts#L262) is not an array when it should.

sudoshreyansh:
**Description**

- replaced `Array.at` function usage with `[]` operator 
- configured Jest to disable polyfilling in test environment

**Related issue(s)**
Fixes #268

### DIFF
--- a/src/lib/glee.ts
+++ b/src/lib/glee.ts
@@ -259,7 +259,7 @@ export default class Glee extends EventEmitter {
   }
 
   private _execErrorMiddleware (emws: ChannelErrorMiddlewareTuple[], index: number, error: Error, message: GleeMessage) {
-    emws.at(index).fn(error, message, (err: Error) => {
+    emws[index].fn(error, message, (err: Error) => {
       if (!emws[index+1]) return
       this._execErrorMiddleware.call(null, emws, index+1, err, message)
     })
--- a/jest.config.js
+++ b/jest.config.js
@@ -4,7 +4,6 @@ export default {
   testEnvironment: 'node',
   setupFilesAfterEnv: ['./jest.setup.ts'],
   collectCoverage: true,
-  setupFiles: ['core-js'],
   silent: false,
   testPathIgnorePatterns: ['functions.test.ts'],
   globals: {


src/lib/glee.ts
```diff
@@ -259,7 +259,7 @@ export default class Glee extends EventEmitter {
   }
 
   private _execErrorMiddleware (emws: ChannelErrorMiddlewareTuple[], index: number, error: Error, message: GleeMessage) {
-    emws.at(index).fn(error, message, (err: Error) => {
+    emws[index as number].fn(error, message, (err: Error) => {
```
fmvilas, line 262: Watch out for negative indices. `Array.at` supports relative indexing, meaning -1 would not fail but instead it will get the last item on the array: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#parameters.

fmvilas (changes requested):
Really cool you found it 👏

src/lib/glee.ts
sudoshreyansh: I don't think there's any way to have a negative `index` with the current code. Should I add a check just in case?

fmvilas: Yeah, as you said, not possible with the **current** code. But code continues to evolve so I'd put a check there just in case.

fmvilas (approved):
LGTM 👍

Issue #256, opened 2022-02-24
Add support for Kafka


fmvilas:
#### Reason/Context

Currently, Glee only supports WebSockets (server) and MQTT protocols. This means it's unusable for those using Kafka, AMQP, or any other different protocol.

#### Description

We should add support for Kafka. In theory, it should be just a matter of adding a new adapter in [the `adapters` folder](https://github.com/asyncapi/glee/tree/master/src/adapters). However, we should also take into account that Kafka supports exactly-once semantics (unlike MQTT and WebSockets) and, therefore, it may be a good idea to address #27 first, although not required.

#### References

* https://www.baeldung.com/kafka-exactly-once
* https://www.confluent.io/blog/exactly-once-semantics-are-possible-heres-how-apache-kafka-does-it/

Ruchip16:
@fmvilas Would like to work on this issue as a part of the Mentorship program

fmvilas:
This is cool! This is an important one! Good choice :)

We'll wait until June 6 to announce the candidates but in the meantime feel free to ask any questions you may have. Good luck! 🤞

Ruchip16:
Sure, hoping for the best meanwhile would check the references attached above to get a more detailed understanding of the issue.

fmvilas:
Congratulations @Ruchip16 for being selected for the AsyncAPI Mentorship program to work on this issue 🎉 I knew it was a good choice 😄 I'll be sending DM to you on Slack so we can start working on it when it's best for both.

Ruchip16:
sureee suree cant wait to dive deep into the issue

fmvilas:
We're working on it dear bot. Don't close it :)

fmvilas:
It's already there! #342 Closing the issue.

PR #342, opened 2022-10-08
feat: kafka adapter 

Ruchip16:
**Description**

Opening a draft PR for issue #256

- Also, I did add a condition for Kafka protocol in the RegisterAdapters but it still results in the same error upon doing `npm run dev` in the `cd examples/dummy` folder in order to check whether our Kafka broker connects to the glee or not in that example

image

**Related issue(s)**


Souvikns:
You have to run `npm run dev` on the root once for the changes to be built and update the `/dist` that the examples will be using. 

Screenshot 2022-10-10 at 3 06 12 PM

asyncapi.yaml
```diff
@@ -0,0 +1,163 @@
...
+    * Dim a specific streetlight 😎
+    * Receive real-time information about environmental lighting conditions 📈
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0
+
+servers:
+  test:
+    url: test.mykafkacluster.org:8092
+    protocol: kafka-secure
```
fmvilas, line 20: For testing, you may want to use `kafka` instead. Otherwise, you'll have to provide certificates to make it "secure".

asyncapi.yaml
```diff
@@ -0,0 +1,163 @@
+asyncapi: '2.4.0'
```
fmvilas, line 1: Please move this example to the `examples` folder or get rid of it. Whatever you prefer :)

package.json
```diff
@@ -36,11 +36,12 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@asyncapi/generator": "^1.9.3",
+    "@asyncapi/glee": "^0.10.23",
```
fmvilas, line 39: You don't need Glee here. Actually, you're installing Glee on Glee 😅 so not sure how this would behave hehe!

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,166 @@
+import fs from 'fs'
+import { Kafka, Consumer } from 'kafkajs'
+import Adapter from '../../lib/adapter.js'
+import GleeMessage from '../../lib/message.js'
+
+
```
fmvilas, line 6:
```suggestion
```

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,166 @@
...
+class KafkaAdapter extends Adapter {
+  private kafka: Kafka
+  private firstConnect: boolean = true
+  name(): string {
+    return 'Kafka adapter'
+  }
+
+  async connect() {
+    const broker = this.AsyncAPIServer.url()
```
fmvilas, line 15: If it's a URL, I'd instead name it as brokerUrl:

```suggestion
    const brokerUrl = this.AsyncAPIServer.url()
```

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,166 @@
...
+          connection: consumer,
+          channels: this.channelNames
+        })
+      }
+    })
+
+    await consumer.connect()
+    
+    const subscribedChannels = this.getSubscribedChannels()
+    await consumer.subscribe({ topics: subscribedChannels, fromBeginning: true })
```
fmvilas, line 36: Not for now but let's not forget to make "fromBeginning" configurable.

src/registerAdapters.ts
```diff
@@ -31,6 +32,12 @@ function registerAdapterForServer(serverName: string, server: Server, app: Glee,
       server,
       parsedAsyncAPI,
     })
+  } else if (['kafka'].includes(protocol)) {
```
fmvilas, line 35: You're not handling the `kafka-secure` protocol yet but you will so let's better add it now:

```suggestion
  } else if (['kafka', 'kafka-secure'].includes(protocol)) {
```

fmvilas (changes requested):
This is a great first approach, @Ruchip16 👏 

Left some comments that I think can have a positive impact. Please, also remove all the package-lock.json changes from this PR. They shouldn't be part of this pull request 🙏 

Keep up the great work!

asyncapi.yaml
Ruchip16: Yea we used this just for testing whether the kafka broker is working or not, we don't need this, so i am deleting it

package.json
Ruchip16: ohkk that happened bymistakely ig when I was trying to run examples in the start😅, but no worries I removed that

package.json
```diff
@@ -50,11 +51,12 @@
     "emojis": "^1.0.10",
     "eslint-plugin-github": "^4.3.5",
     "eslint-plugin-security": "^1.4.0",
+    "kafkajs": "^2.2.0",
```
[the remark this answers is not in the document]
Ruchip16: done, I have reverted the changes ig these changes happen automatically when I install the `better-ajv-errors` module because I was getting errors on doing npm run dev in these modules, no worries I reverted them back to the original versions

src/adapters/kafka/index.ts
Ruchip16: yeaa definitely

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,165 @@
...
+  //         channels: this.channelNames,
+  //       })
+  //     })
+
+  //     this.kafka.on('error', (error) => {
+  //       this.emit('error', error)
+  //     })
+  // }
+
+  _send(message: GleeMessage): Promise<void> {
```
fmvilas, line 120: @Ruchip16 do you need help implementing this part of the binding?

Ruchip16: I am currently implementing that as per my understanding, I would push the changes once I am done and then we can have a meeting to discuss them, wyt?

fmvilas: Absolutely 👍 We can also discuss them here before the meeting, so everyone is involved in the discussion.

Ruchip16: donee

fmvilas (changes requested):
In general, I think your experimentation is going great. You have a big mess in the adapter file, so I recommend you clean it out and start slowly adding things that you are sure should be there. No commented code, no copy-pasted code, clean it all and start over.

src/adapters/kafka/index.ts
```diff
@@ -65,133 +41,22 @@ class KafkaAdapter extends Adapter {
...
-          connection: this.kafka,
-          channels: this.channelNames,
-        })
-      })
-
-      this.kafka.on('error', (error) => {
-        this.emit('error', error)
-      })
+    run().catch(console.error)
+    await consumer.disconnect()
```
fmvilas, line 58: If you call `disconnect`, it's going to only consume one message and then stop. We shouldn't ever disconnect from the broker.

src/adapters/kafka/index.ts
```diff
@@ -65,133 +41,22 @@ class KafkaAdapter extends Adapter {
...
+      await consumer.connect() 
+      const subscribedChannels = this.getSubscribedChannels()
+      await consumer.subscribe({ topics: subscribedChannels, fromBeginning: true })
+      await consumer.run({
+        eachMessage: async ({ topic, partition, message }) => {
+          const prefix = `${topic}[${partition} | ${message.offset}] / ${message.timestamp}`
+          console.log(`- ${prefix} ${message.key}#${message.value}`)
+        },
       })
+    }
```
fmvilas, line 55: You don't need to wrap this code inside a `run` function, do you? It should work just fine without it.

src/adapters/kafka/index.ts
```diff
@@ -65,133 +41,22 @@ class KafkaAdapter extends Adapter {
...
-        this.emit('close', {
-          connection: this.kafka,
-          channels: this.channelNames,
-        })
-      })
-
-      this.kafka.on('error', (error) => {
-        this.emit('error', error)
-      })
+    run().catch(console.error)
```
fmvilas, line 57: You also don't need this. The `connect` function of this adapter is already running. You don't need one more function to run inside.

src/adapters/kafka/index.ts
```diff
@@ -13,47 +12,24 @@ class KafkaAdapter extends Adapter {
...
 
+    const producer = kafka.producer()
     await producer.connect()
     await producer.send({
       topic: 'test-topic',
       messages: [
         { value: 'Hello KafkaJS user!' },
       ],
     })
+    await producer.disconnect()
```
fmvilas, line 30: You should not have a producer here. Otherwise, it will send a message when you connect to the broker.

fmvilas (changes requested):

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,63 @@
...
+          // value: message.value.toString(),
+          // key: message.key,
+          // offset: message.offset,
+          // timestamp: message.timestamp,
+          // headers: message.headers,
+        });
+        const prefix = `${topic}[${partition} | ${message.offset}] / ${message.timestamp}`
+        console.log(`- ${prefix} ${message.key}#${message.value}`)
+
+        gleemessage.send();
```
fmvilas, line 57: Glee uses an internal event system. So you don't tell Glee directly to do anything with the message but instead, you trigger an event. Specifically, the `message` event. You can learn more from Node.js event emitters here: https://nodejs.dev/en/learn/the-nodejs-event-emitter/.

```suggestion
        this.emit('message', gleeMessage, consumer)
```

`gleeMessage` is the GleeMessage instance we just created above.
`consumer` is the object that contains the connection to the Kafka broker.

Glee is listening for the `message` event and, when it happens, it starts processing the given message (`gleeMessage`) for the given connection (`consumer`).

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      ssl: true,
+      sasl: {
+        mechanism: 'plain',
+        username: process.env.KAFKA_USERNAME,
+        password: process.env.KAFKA_PASSWORD,
+      },
+    })
+
+    const consumer = kafka.consumer({ groupId: 'glee-group' })   // groupID: hardcoded need to change afterwards
+    const producer = kafka.producer();
```
fmvilas, line 27: You should not have a producer here. We're just connecting to subscribe at this moment.

```suggestion
```

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,63 @@
...
+          adapter: this,
+          connection: consumer,
+          channels: this.channelNames
+        })
+      }
+    })
+    await consumer.connect() 
+    const subscribedChannels = this.getSubscribedChannels()
+    await consumer.subscribe({ topics: subscribedChannels, fromBeginning: true })
+    await producer.connect();
```
fmvilas, line 42: You should not have a producer here. We're just connecting to subscribe at this moment.

```suggestion
```

fmvilas (changes requested):
You're close to having a subscriber ready! Keep pushing! 💪

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      eachMessage: async ({ topic, partition, message }) => {
+        const gleemessage = new GleeMessage({
+          // topic: topic,
+          // partition: partition,
+          // value: message.value.toString(),
+          // key: message.key,
+          // offset: message.offset,
+          // timestamp: message.timestamp,
+          // headers: message.headers,
+        });
```
fmvilas, line 53: Following our offline conversation. This is giving you an error because the GleeMessage constructor expects a different set of values: https://github.com/asyncapi/glee/blob/master/src/lib/message.ts#L6.

I'd group things the following way:
  * GleeMessage channel: topic.
  * GleeMessage payload: message.value. Don't call `toString()` here. Glee is capable of handling it.
  * GleeMessage headers: `partition`, `message.key`, `message.offset`, `message.timestamp`, and `message.headers`. If there's anything else in the `message` object, it should most probably be in the GleeMessage headers too.

So here's how you'd do it (beware there may be errors as I'm not testing it):

```suggestion
        const gleeMessage = new GleeMessage({
          channel: topic,
          payload: message.value,
          headers: {
            partition: partition,
            key: message.key,
            offset: message.offset,
            timestamp: message.timestamp,
            ...message.headers,
          }
        })
```

fmvilas (changes requested):
Sorry, for some reason this suggestion wasn't added before 🤔

examples/dummy/asyncapi.yaml
```diff
@@ -4,6 +4,9 @@ info:
   version: 1.0.0
   description: This service is in charge of processing user signups
 servers:
+  mykafka:
+    url: kafka://localhost:9092
+    protocol: kafka
```
fmvilas, line 9: If you're gonna use Confluent instead of a local Kafka broker, this URL should be pointing to your Confluent Kafka broker and the protocol should be `kafka-secure`. You don't have to push the code changes but that's how you should be testing it locally.

src/adapters/kafka/index.ts
```diff
@@ -42,21 +44,15 @@ class KafkaAdapter extends Adapter {
...
-          // topic: topic,
-          // partition: partition,
-          // value: message.value.toString(),
-          // key: message.key,
-          // offset: message.offset,
-          // timestamp: message.timestamp,
-          // headers: message.headers,
-        });
         const prefix = `${topic}[${partition} | ${message.offset}] / ${message.timestamp}`
         console.log(`- ${prefix} ${message.key}#${message.value}`)
```
fmvilas, line 48: Here you have to create the GleeMessage and pass it over to Glee:

```suggestion
        const gleeMessage = this.createMessage(message)
        this.emit('message', gleeMessage)
```

Then you need to add another function called `createMessage` to this adapter. Similar to the one here: https://github.com/asyncapi/glee/blob/master/src/adapters/mqtt/index.ts#L151. The difference is that instead of getting an MQTT `packet` you get a Kafka `message`. You have to convert the message to a GleeMessage and return it. I think we discussed how to convert to GleeMessage on Slack.

fmvilas (changes requested):

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,58 @@
...
+        })
+      }
+    })
+    await consumer.connect() 
+    const subscribedChannels = this.getSubscribedChannels()
+    await consumer.subscribe({ topics: subscribedChannels, fromBeginning: true })
+    await consumer.run({
+      eachMessage: async ({ topic, partition, message }) => {
+        const prefix = `${topic}[${partition} | ${message.offset}] / ${message.timestamp}`
+        console.log(`- ${prefix} ${message.key}#${message.value}`)
```
fmvilas, line 47: Here you have to create the GleeMessage and pass it over to Glee:

```suggestion
        const gleeMessage = this.createMessage(message)
        this.emit('message', gleeMessage)
```

Then you need to add another function called `createMessage` to this adapter. Similar to the one here: https://github.com/asyncapi/glee/blob/master/src/adapters/mqtt/index.ts#L151. The difference is that instead of getting an MQTT `packet` you get a Kafka `message`. You have to convert the message to a GleeMessage and return it. I think we discussed how to convert to GleeMessage on Slack.

src/lib/glee.ts
```diff
@@ -27,6 +27,12 @@ type ClusterAdapterRecord = {
 }
 
 export default class Glee extends EventEmitter {
+  static createMessage(value: any) {
+    throw new Error('Method not implemented.')
+  }
+  private _createMessage(value: any) {
+    throw new Error('Method not implemented.')
+  }
```
fmvilas, line 35: This should not be here. You don't need to touch this file to add your adapter :)

```suggestion
```

src/adapters/kafka/index.ts
fmvilas: Actually, the way to convert a Kafka message to a GleeMessage is defined here: https://github.com/asyncapi/glee/pull/342#discussion_r1041781088.

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,67 @@
...
+          channels: this.channelNames
+        })
+      }
+    })
+    await consumer.connect() 
+    const subscribedChannels = this.getSubscribedChannels()
+    await consumer.subscribe({ topics: subscribedChannels, fromBeginning: true })
+    await consumer.run({
+      eachMessage: async ({ topic, partition, message }) => {
+        const msg = this._createMessage(message.value)
```
fmvilas, line 45: The `_createMessage` function is expecting a topic and a message, not just the message value:

```suggestion
        const msg = this._createMessage(topic, message)
```

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,67 @@
...
+      },
+    }) 
+  }
+
+  _createMessage(message) {
+    const gleeMessage = new GleeMessage({
+      payload: message.value,
+      headers: {
+        'x-kafka-partition': message.partition,
+        'x-kafka-offset': message.offset,
```
fmvilas, line 56: There's no need to prepend `x-kafka-` to these headers and you need to add the rest of the headers in the message:

```suggestion
        partition: message.partition,
        offset: message.offset,
        ...message.headers,
```

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,67 @@
...
+      eachMessage: async ({ topic, partition, message }) => {
+        const msg = this._createMessage(message.value)
+        this.emit('message', msg)
+      },
+    }) 
+  }
+
+  _createMessage(message) {
+    const gleeMessage = new GleeMessage({
+      payload: message.value,
```
fmvilas, line 53: You can return directly here and this way you don't have to return below + you need to add the channel/topic + you need to pass topic to the _createMessage function, otherwise you don't have this information:

```suggestion
  _createMessage(topic, message) {
    return new GleeMessage({
      channel: topic,
      payload: message.value,
```

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,67 @@
...
+      headers: {
+        'x-kafka-partition': message.partition,
+        'x-kafka-offset': message.offset,
+      },
+    })
+
+    return new GleeMessage({
+      payload: message.payload,
+      channel: message.topic,
+    })
```
fmvilas, line 63: If you apply the suggestions above you don't need this anymore:

```suggestion
```

fmvilas (changes requested):

src/lib/glee.ts
Ruchip16: noo i am not touching these files should i remove this lines from 30 to 35?

fmvilas: Yes, remove these lines. It's not needed.

src/adapters/kafka/index.ts
Ruchip16: `...message.headers`  is supposed to be like this or does it imply that there are more headers that needs to be added?

fmvilas: The `message.headers` object may contain other headers that we can not know in advance. With `...message.headers` we tell JS/TS to include all of them in this object. The `...` is the [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax).

examples/dummy - Copy/.env
```diff
@@ -0,0 +1,2 @@
+GLEE_SERVER_NAMES=mykafka
+GLEE_SERVER_CERTS=mosquitto:mosquitto.org.crt
```
fmvilas, line 2: Remove this as it's not needed:

```suggestion
```

examples/dummy - Copy/customServer.js
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable no-undef, security/detect-non-literal-fs-filename */
```
fmvilas, line 1: Remove this whole file, it's not needed.

examples/dummy - Copy/glee.config.js
```diff
@@ -0,0 +1,15 @@
+// import customServer from './customServer.js'
```
fmvilas, line 1: Remove this whole file.

examples/dummy - Copy/index.html
```diff
@@ -0,0 +1,15 @@
+<script>
```
fmvilas, line 1: Remove this file.

examples/dummy - Copy/lifecycle/announceServer.ts
```diff
@@ -0,0 +1,14 @@
+export default async function () {
```
fmvilas, line 1: Remove this file and the whole "lifecycle" folder.

examples/dummy - Copy/mosquitto.org.crt
```diff
@@ -0,0 +1,24 @@
+-----BEGIN CERTIFICATE-----
```
fmvilas, line 1: Remove this file.

examples/dummy - Copy/socket.io.html
```diff
@@ -0,0 +1,16 @@
+<script src="https://cdn.socket.io/3.1.3/socket.io.min.js"
```
fmvilas, line 1: Remove this file.

fmvilas (changes requested):

examples/dummy/.env
```diff
@@ -1 +1,2 @@
+GLEE_SERVER_NAMES=mykafka
```
fmvilas, line 1: You have to revert this value to the original one.

.gitignore
```diff
@@ -2,4 +2,5 @@ node_modules
 .DS_Store
 dist
 coverage
-.vscode
\ No newline at end of file
+.vscode
+local.env
```
fmvilas, line 6: Don't put this here. Put it in a .gitignore file in the kafka-test example.

examples/dummy/asyncapi.yaml
```diff
@@ -4,6 +4,9 @@ info:
   version: 1.0.0
   description: This service is in charge of processing user signups
 servers:
+  mykafka:
```
fmvilas, line 7: Remove this server from the dummy example now that you have kafka-test.

fmvilas (changes requested):
Let's go! This is probably working already, right?

examples/kafka-test/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "glee-example",
+  "version": "0.1.0",
+  "description": "An example of the AsyncAPI Glee framework",
+  "type": "module",
+  "scripts": {
+    "dev": "glee dev",
+    "start": "glee start"
+  },
+  "author": "Fran Mendez",
```
fmvilas, line 10: 🙂 

```suggestion
  "author": "Ruchi Pakhle",
```

src/adapters/kafka/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
+version: '3'
```
fmvilas, line 1: Now you can move this file to `examples/kafka-test`.

fmvilas (changes requested):

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    const consumer = kafka.consumer({ groupId: 'glee-group' })   // groupID: hardcoded need to change afterwards
+  
+    consumer.on('consumer.connect', () => {
+      if (this.firstConnect) {
+        this.firstConnect = false
+        this.emit('connect', {
+          name: this.name(),
+          adapter: this,
+          connection: consumer,
+          channels: this.channelNames
```
fmvilas, line 36: This is making subscription fail:

```suggestion
          channels: this.getSubscribedChannels()
```

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,66 @@
...
+        })
+      }
+    })
+    await consumer.connect() 
+    const subscribedChannels = this.getSubscribedChannels()
+    await consumer.subscribe({ topics: subscribedChannels, fromBeginning: true })
+    await consumer.run({
+      eachMessage: async ({ topic, partition, message }) => {
+        const msg = this._createMessage(topic, message)
+        this.emit('message', msg)
```
fmvilas, line 46: You need to pass the consumer in the second parameter:

```suggestion
        this.emit('message', msg, consumer)
```

fmvilas (changes requested):

examples/kafka-test/package.json
Ruchip16: you are the OG author fran, anyday 😄

examples/kafka-test/functions/onTest.ts
```diff
@@ -0,0 +1,3 @@
+export default async function (event) {
+  console.log(event)
```
fmvilas, line 2: Let's implement an echo functionality. Let's send back exactly the same message you receive. This will help you test the "send" functionality. After sending a message from the Confluent UI, you should see a message appear in your topic with the same details. Hopefully, it doesn't enter an endless loop 😄 If it does, you can just create another topic where the "echo" will be sent to.

```suggestion
  return {
    send: [{
      server: event.serverName,
      channel: event.channel,
      payload: event.payload,
      headers: event.headers,
    }]
  }
```

fmvilas (commented):
Alright, so you got the subscribe functionality working 👏 Left a comment on how to test the "send" functionality. Let me know if you have questions.

examples/kafka-test/package.json
fmvilas: haha! That's not true. You're the author of this example so feel free to change it!

.gitignore
fmvilas: No, leave .vscode there. Otherwise, we have to put it in every example. Remember `local.env` is just for you now while developing.

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      payload: message.value,
+      headers: {
+        partition: message.partition,
+        key: message.key,
+        offset: message.offset,
+        timestamp: message.timestamp,
+        ...message.headers,
+      },
+    })
+  }
```
fmvilas, line 63: When I tested the code, I realized `partition` was always "undefined". We should pass `partition` to `this._createMessage()` in order to assign it properly:

```suggestion
      eachMessage: async ({ topic, partition, message }) => {
        const msg = this._createMessage(topic, partition, message)
        this.emit('message', msg, consumer)
      },
    }) 
  }

  _createMessage(topic, partition, message) {
    return new GleeMessage({
      channel: topic,
      payload: message.value,
      headers: {
        partition,
        key: message.key,
        offset: message.offset,
        timestamp: message.timestamp,
        ...message.headers,
      },
    })
  }
```

fmvilas (changes requested):

src/adapters/kafka/index.ts
```diff
@@ -42,18 +40,29 @@ class KafkaAdapter extends Adapter {
...
       },
-    }) 
+    })
+    
+    const producer = kafka.producer()
+    await producer.connect()
+    await producer.send({
+      topic: 'glee-topic',
+      messages: [{ value: 'Hello, Glee!' }],
+    })
```
fmvilas, line 53: If you put the producer code here, it will send a "Hello, Glee" message to Kafka every time we connect to the broker. And that's not what we want, right? We want to send a message (GleeMessage) to Kafka every time someone calls the "send" function below. So you have to move your code over there.

src/adapters/kafka/index.ts
```diff
@@ -42,18 +40,29 @@ class KafkaAdapter extends Adapter {
...
+  }
+
+  async send(message: GleeMessage) {
+
```
fmvilas, line 57:
```suggestion
    const producer = kafka.producer()
    await producer.connect()
    await producer.send({
      topic: message.channel,
      messages: [{
        value: message.payload,
      }],
    })
```

You can pass more details to the message object above. The structure documentation is here: https://kafka.js.org/docs/producing#message-structure.

In `_createMessage`, you converted a Kafka message to a GleeMessage. Here you have to do the opposite. You get a GleeMessage and have to convert it to a Kafka message.

You probably have to call `await producer.disconnect()` after sending the message.

fmvilas (changes requested):

src/adapters/kafka/index.ts
Ruchip16: yess i actually put the producer part in the send function only but then the kafka client is declared over here and that was not accessible in the send function below, but i guess i have to declare that globally so that its accessible to each function

src/lib/message.ts
```diff
@@ -20,6 +20,7 @@ interface IReply {
 }
 
 class GleeMessage extends EventEmitter {
+  [x: string]: any
```
fmvilas, line 23: Don't change this code.

src/adapters/kafka/index.ts
```diff
@@ -44,17 +43,22 @@ class KafkaAdapter extends Adapter {
         this.emit('message', msg, consumer)
       },
     })
-    
+  }
+
+  async send(message: GleeMessage) {
     const producer = kafka.producer()
```
fmvilas, line 49: `kafka` is not defined here. It's in the `connect` function. My suggestion is that, in the `connect` function, you assign it to `this.kafka` instead of doing `const kafka`. This way, the variable will be available on the whole class using `this.kafka`.

fmvilas (changes requested):

examples/kafka-test/asyncapi.yaml
```diff
@@ -0,0 +1,18 @@
...
+channels:
+  test:
+    publish:
+      operationId: onTest
+      message:
+        payload:
+          type: object
+          properties:
+            test:
+              type: string
```
fmvilas, line 18: Your `onTest` function is sending a message to the topic `produce` so that should be defined in the AsyncAPI file:

```suggestion
        $ref: '#/components/messages/testMessage'
  produce:
    subscribe:
      message:
        $ref: '#/components/messages/testMessage'
components:
  messages:
    testMessage:
      payload:
        type: object
        properties:
          test:
            type: string
```

Otherwise, you'll get the following error:

Captura de Pantalla 2022-12-22 a las 19 18 53

fmvilas (changes requested):
The "send" functionality is working great. Just adapt the AsyncAPI file and you'll see it working perfectly :)

src/adapters/kafka/index.ts
```diff
@@ -14,16 +15,32 @@ class KafkaAdapter extends Adapter {
     this.kafka = new Kafka({
       clientId: 'glee-app',  // clientID: hardcoded need to change afterwards 
       brokers: [brokerUrl.host],
-      ssl: true,
+      ssl: {
```
Ruchip16, line 18: i am just testing stuff and trying to putting it over here, also in SSL we need all these properties right?
Also how do would we get the values that needed to be passed like `client.key, client.pem`

src/adapters/kafka/index.ts
```diff
@@ -14,16 +15,32 @@ class KafkaAdapter extends Adapter {
...
+      const secName = Object.keys(sec.json())[0]
+      return this.parsedAsyncAPI.components().securityScheme(secName)
+    })
+    const userAndPasswordSecurityReq = securityRequirements.find(sec => sec.type() === 'userPassword')
+    const X509SecurityReq = securityRequirements.find(sec => sec.type() === 'X509')
+    const url = new URL(this.AsyncAPIServer.url())
 
+    const certsConfig = process.env.GLEE_SERVER_CERTS?.split(',').map(t => t.split(':'))
+    const certs = certsConfig?.filter(tuple => tuple[0] === this.serverName)?.map(t => fs.readFileSync(t[1]))
+
```
Ruchip16, line 41: I referred to this from the MQTT adapter and was not sure if we should put things in a similar way or how we should put it, so do we have to do this in a similar manner in our adapter as well or everything would get passed in SSL and SASL?
@fmvilas

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,95 @@
...
+        username: process.env.KAFKA_USERNAME,
+        password: process.env.KAFKA_PASSWORD,
+      },
+    })
+
+    const securityRequirements = (this.AsyncAPIServer.security() || []).map(sec => {
+      const secName = Object.keys(sec.json())[0]
+      return this.parsedAsyncAPI.components().securityScheme(secName)
+    })
+    const userAndPasswordSecurityReq = securityRequirements.find(sec => sec.type() === 'userPassword')
```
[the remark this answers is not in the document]
fmvilas: Let's just do the same for `scramSha256` and `scramSha512` types. In such a case, the only difference with the code above is that you have to replace `plain` with `scramSha256` or `scramSha512` respectively. That would be more than enough for this PR 👍

src/adapters/kafka/index.ts
fmvilas: You need to get these values before line 15 because you need to pass them to the SASL config (forget SSL, it has nothing to do with the AsyncAPI file but with the glee.config.js file). I hope that makes sense.

fmvilas (changes requested):

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,95 @@
...
+    const X509SecurityReq = securityRequirements.find(sec => sec.type() === 'X509')
+    const url = new URL(this.AsyncAPIServer.url())
+
+    const certsConfig = process.env.GLEE_SERVER_CERTS?.split(',').map(t => t.split(':'))
+    const certs = certsConfig?.filter(tuple => tuple[0] === this.serverName)?.map(t => fs.readFileSync(t[1]))
```
fmvilas, line 40: Oh, forgot to say that you don't need this. Kafka is not supporting X509 certifcates AFAIK.

src/lib/message.ts
```diff
@@ -30,6 +30,8 @@ class GleeMessage extends EventEmitter {
   private _outbound: boolean
   private _cluster: boolean
   private _params: { [key: string]: string }
+  timestamp: string
+  key: string | Buffer
```
fmvilas, line 34: You have to get rid of these two lines. They don't belong here.

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    this.kafka = new Kafka({
+      clientId: 'glee-app',  // clientID: hardcoded need to change afterwards 
+      brokers: [brokerUrl.host],
+      ssl: {
+        rejectUnauthorized: true,
+        key: fs.readFileSync('client.key'),
+        cert: undefined
+      },
+      sasl: {
+        mechanism: 'plain',
```
fmvilas, line 25: You have to guess if it's going to be `plain`, `scramSha256`, or `scramSha512`. You have this info in the AsyncAPI file. If it's not specified, it's ok to assign the whole `sasl` object here to `undefined`.

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,87 @@
...
+        username: process.env.KAFKA_USERNAME
+        ? kafkaOptions?.authentication?.userPassword.username
```
fmvilas, line 27:
```suggestion
        ? kafkaOptions?.authentication?.userPassword?.username
```

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,87 @@
...
+        : undefined,
+        password: process.env.KAFKA_PASSWORD
+        ? kafkaOptions?.authentication?.userPassword.username
```
fmvilas, line 30:
```suggestion
        ? kafkaOptions?.authentication?.userPassword?.username
```

fmvilas (changes requested):
Alright getting there!!

src/lib/message.ts
Ruchip16: but then in the index.ts file it gives error saying that key & timestamp are not defined in GleeMessage

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  async send(message: GleeMessage) {
+    const producer = this.kafka.producer()
+    await producer.connect()
+    await producer.send({
+      topic: message.channel,
+      messages: [{
+        key: message.key,
+        value: message.payload,
+        timestamp: message.timestamp,
```
fmvilas, line 66:
```suggestion
        key: message.headers.key,
        value: message.payload,
        timestamp: message.headers.timestamp,
```

fmvilas (changes requested):

src/lib/message.ts
fmvilas: True but that was just a typo you made there. `key` and `timestamp` are not properties of the message object but of the headers. Left another suggestion below.

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,96 @@
...
+        cert: undefined
+      },
+      sasl: {
+        mechanism: 'scram-sha-512',
+        username: process.env.KAFKA_USERNAME
+        ? kafkaOptions?.authentication?.userPassword?.username
+        : undefined,
+        password: process.env.KAFKA_PASSWORD
+        ? kafkaOptions?.authentication?.userPassword?.username
+        : undefined,
```
fmvilas, line 40: You don't have to check environment variables now. If the user wants to use environment variables, they can do it in the config file. The only thing you have to do now is to check the config file values:
```suggestion
        username: kafkaOptions?.authentication?.userPassword?.username,
        password: kafkaOptions?.authentication?.userPassword?.username,
```

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    this.kafka = new Kafka({
+      clientId: 'glee-app',  // clientID: hardcoded need to change afterwards 
+      brokers: [brokerUrl.host],
+      ssl: {
+        rejectUnauthorized: true,
+        key: fs.readFileSync('client.key'),
+        cert: undefined
+      },
+      sasl: {
+        mechanism: 'scram-sha-512',
```
fmvilas, line 34: Alright so here is where I meant you should be guessing the mechanism. On line 21 above you do:

```js
const userAndPasswordSecurityReq = securityRequirements.find(
  (sec) => sec.type() === 'userPassword'
)
```

This is the same as saying: "please find a security requirement in the AsyncAPI file that has the type `userPassword`". If `userAndPasswordSecurityReq` is different than `undefined`, it means it found it. In such a case, `mechanism` in this line should be `plain`.

Now, you're missing other checks there. You also have to find if there are any security requirements of the type `scramSha256` or `scramSha512`, so you have to find them the same way:

```js
const scramSha256SecurityReq = securityRequirements.find(
  (sec) => sec.type() === 'scramSha256'
)
const scramSha512SecurityReq = securityRequirements.find(
  (sec) => sec.type() === 'scramSha512'
)
```

If `scramSha256SecurityReq` is different than `undefined` then `mechanism` should be `scramSha256`. If `scramSha512SecurityReq` is different than `undefined` then `mechanism` should be `scramSha512`. If all the 3 are `undefined` or the `securityRequirements` variable (on line 15) is `undefined` then `mechanism` should be `plain`.

I left an example above about how you can add a scramSha256 security requirement in your AsyncAPI file.

fmvilas (changes requested):

src/adapters/kafka/index.ts
Ruchip16: ok, got it, so its basically one security type from user password, scramsha256, scramsha512 and update it in YAML file

fmvilas: I've added a suggestion that you didn't apply yet to make the security mechanism of our example of type `scramSha256`. And yeah, if no mechanism is defined, it should be `plain` too.

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,99 @@
...
+    const scramSha512SecurityReq = securityRequirements.find(
+      (sec) => sec.type() === 'scramSha512'
+    )
+  
+    const brokerUrl = new URL(this.AsyncAPIServer.url())
+    this.kafka = new Kafka({
+      clientId: 'glee-app', 
+      brokers: [brokerUrl.host],
+      ssl: {
+        rejectUnauthorized: true,
```
fmvilas, line 35:
```suggestion
        rejectUnauthorized: kafkaOptions?.authentication?.rejectUnauthorized,
```

.gitignore
```diff
@@ -4,4 +4,3 @@ dist
 coverage
 .vscode
 .glee
-.next
```
fmvilas, line 7: I think this is another accidental removal. Please add it back: https://github.com/asyncapi/glee/blob/master/.gitignore

examples/kafka-test/functions/onTest.ts
```diff
@@ -0,0 +1,10 @@
+export default async function (event) {
+  return {
+    send: [{
+      server: event.serverName,
+      channel: "produce",
```
fmvilas, line 5:
```suggestion
      channel: 'produce',
```

examples/kafka-test/local.env
```diff
@@ -0,0 +1,2 @@
+KAFKA_USERNAME=DBM6YJZBTYHUCVAR
```
fmvilas, line 1: You can remove this file already.

src/lib/functions.ts
```diff
@@ -22,7 +22,6 @@ const OutboundMessageSchema = {
     headers: {
       type: 'object',
       propertyNames: { type: 'string' },
-      additionalProperties: { type: 'string' }
```
fmvilas, line 25: This change shouldn't be here. Please add this line back.

src/index.ts
```diff
@@ -22,7 +22,7 @@ import { getSelectedServerNames } from './lib/servers.js'
 import { EnrichedEvent } from './lib/adapter.js'
 import { ClusterEvent } from './lib/cluster.js'
 
-dotenvExpand(dotenv.config())
+dotenvExpand(dotenv.config({ path: 'local.env' }))
```
fmvilas, line 25: You can already remove this change.

src/lib/message.ts
```diff
@@ -30,7 +30,6 @@ class GleeMessage extends EventEmitter {
   private _outbound: boolean
   private _cluster: boolean
   private _params: { [key: string]: string }
-
```
fmvilas, line 33: This change shouldn't be here. Please add this line back.

src/registerAdapters.ts
```diff
@@ -33,10 +34,16 @@ function registerAdapterForServer(serverName: string, server: Server, app: Glee,
...
+    app.addAdapter(KafkaAdapter, {
+      serverName,
+      server,
+      parsedAsyncAPI,
+    })
   } else if (['amqp', 'amqps'].includes(protocol)) {
     // TODO: Implement AMQP support
   } else if (['ws', 'wss'].includes(protocol)) {
-    const configWsAdapter = config?.websocket?.server?.adapter
+    const configWsAdapter = config?.websocket?.server.adapter
```
fmvilas, line 46: This change shouldn't be here. Please add the `?` symbol back.

fmvilas (changes requested):
Alright, made a full review. It's looking good. Just apply these changes and we're done :)

docs/config-file.md
```diff
@@ -68,8 +74,17 @@ export default async function () {
...
 
-Every protocol has different configuration needs so each protocol has unique configurations:
+These configurations apply to Glee itself, rather than any specific protocol.
+
+|Field|Default|Description|
+|--|--|--|
+|glee.gleeDir|`.glee`|Sets the Glee directory. Your sources will be compiled here.|
+|glee.lifecycleDir|`lifecycle`|Path to the directory that stores your [lifecycle events](./lifecycle-events.md).|
+|glee.functionsDir|`functions`| Path to the directory that stores your [functions](./functions.md).|
+|glee.asyncapiFilePath|`asyncapi.(yaml \| yml \| json)`| Path to your AsyncAPI file. |
```
fmvilas, line 87: These changes shouldn't be on your PR. Please merge the latest `master` into your branch (no rebase! 😄).

docs/config-file.md
```diff
@@ -110,3 +125,18 @@ Every protocol has different configuration needs so each protocol has unique con
...
+|kafka.authentication| Kafka authentication configuration|
+|kafka.authentication.key | Kafka Broker Key
+|kafka.authentication.cert| Client certificate
+|kafka.authentication.clientId| Kafka client Id for authentication
+|kafka.authentication.rejectUnauthorized | Boolean flag for accepting the valid SSL certificates
+|kafka.authentication.userPassword| username and password parameters for authentication|
+|kafka.authentication.userPassword.username| username parameter
+|kafka.authentication.userPassword.password| password parameter
+|kafka.authentication.scramSha256.mechanism | ScramSha256 SASL mechanism type for authentication
+|kafka.authentication.scramSha512.mechanism | ScramSha512 SASL mechanism type for authentication
```
fmvilas, line 142:
```suggestion
|kafka.authentication.username| The username to use during authentication.
|kafka.authentication.password| The password to use during authentication.
```

Let's simplify username and password and remove the scram mechanism as it's coming from the AsyncAPI file, not from the config file.

examples/kafka-test/glee.config.js
```diff
@@ -0,0 +1,9 @@
+export default async function () {
+  return {
+    kafkaOptions: {
+        authentication: {
+            keyFilePath: 'XIL3PSQNLE3ROP7A',
```
fmvilas, line 5: Documentation says this should be `key` instead of `keyFilePath`. Let's rename this to `key` because it's clear this is not a file path.

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,99 @@
...
+      (sec) => sec.type() === 'scramSha512'
+    )
+  
+    const brokerUrl = new URL(this.AsyncAPIServer.url())
+    this.kafka = new Kafka({
+      clientId: 'glee-app', 
+      brokers: [brokerUrl.host],
+      ssl: {
+        rejectUnauthorized: kafkaOptions?.authentication?.rejectUnauthorized,
+        key: kafkaOptions?.authentication?.keyFilePath,
```
fmvilas, line 36: As pointed above, this should be `key` instead of `keyFilePath`:

```suggestion
        key: kafkaOptions?.authentication?.key,
```

src/adapters/kafka/index.ts
```diff
@@ -0,0 +1,99 @@
...
+        cert: kafkaOptions?.authentication?.cert,
+      },
+      sasl: {
+        mechanism: (scramSha256SecurityReq? kafkaOptions?.authentication?.scramSha256?.mechanism : undefined) || 
+                   (scramSha512SecurityReq? kafkaOptions?.authentication?.scramSha512?.mechanism : undefined),
+        username: userAndPasswordSecurityReq? kafkaOptions?.authentication?.userPassword?.username : undefined,
+        password: userAndPasswordSecurityReq? kafkaOptions?.authentication?.userPassword?.password : undefined,
```
fmvilas, line 43:
```suggestion
        mechanism: (scramSha256SecurityReq? 'scram-sha-256 : undefined) || 
          (scramSha512SecurityReq? 'scram-sha-512' : undefined) ||
          'plain',
        username: kafkaOptions?.authentication?.username,
        password: kafkaOptions?.authentication?.password,
```

A few things here:

1. If `scramSha256SecurityReq` is not `undefined`, it means it has to be used. No need to go to the config file. The AsyncAPI is the config file in this case. Same for `scramSha512SecurityReq`.
2. If another mechanism is specified, the `plain` should be used.
3. All mechanisms let you pass a username and password so let's not make it dependent on the security requirement.

fmvilas (changes requested):
One more round. Let's go! It's close!

src/adapters/kafka/index.ts
Ruchip16:
```
mechanism: (scramSha256SecurityReq? 'scram-sha-256 : undefined) || 
          (scramSha512SecurityReq? 'scram-sha-512' : undefined) ||
          'plain', 
```
          but this is giving an error saying : Type '{ mechanism: "plain" | "scram-sha-256" | "scram-sha-512"; username: any; password: any; }' is not assignable to type 'SASLOptions | Mechanism'.
  Object literal may only specify known properties, and 'username' does not exist in type 'Mechanism'.ts(2322)
index.d.ts(53, 3): The expected type comes from property 'sasl' which is declared here on type 'KafkaConfig

fmvilas: Pushed a fix for this :)

fmvilas (approved):
LET'S GOOOO!

fmvilas (approved):
NOW YES! LET'S GOOOO 🚀

fmvilas:
Congratulations on finishing your mentorship program, Ruchi! 👏

Ruchip16:
> Congratulations on finishing your mentorship program, Ruchi! 👏

Thank youuu ✨

PR #351, opened 2022-10-24
feat: update glee config to support multiple protocols

### LINKED ISSUE #347: Improve glee config file to support protocol specific authtication
#### Reason/Context
The current `glee.config` file supports only custom WebSocket servers, it would be nice if it could also help with authentication and passing in environment variables. This feature would help #319 pass in authentication details and find a better way to pass in authentication variables for the current `mqtt` adapter. https://github.com/asyncapi/glee/blob/845db8983b19699d6c5277af0ea75fe5d8aab5e5/src/adapters/mqtt/index.ts#L58-L59

#### Description
I will try to propose some ways we can modify the current `glee.config.js` file. 

- `glee.conig.js` should export an async function that would return an object containing protocol-specific configurations
```js
  return {
    websocket: {},
    mqtt: {},
    http: {}, // in future when we support http protocol
    amqp: {} 
  }
}
```
- Since all the protocols have different needs all protocols will have different object structures, which will be documented. 
- In future we will support more protocols and the protocol objets will change too. 
- Every adapter will have access to it's protocol specific config. How it uses this config is up to the adapter. 
- To specify the authentication we will use [security schema](https://github.com/asyncapi/spec/blob/v2.5.0/spec/asyncapi.md#securitySchemeObject) to define rules for authentication. 

#### Websockets 

```js
export default function () {
  return {
    websocket: {
      server: {
        httpServer: customServer,
        adapter: 'native',
        port: process.env.PORT
      },
      client: {
        authentication: {
          bearer: process.env.TOKEN // bearer, api_key, httpApiKey as per the security scheme definition
        }
      }
    }
  }
}
```

#### mqtt
```js
export default function () {
  return {
    mqtt: {
      authentication: {
      username: process.env.USERNAME,
      password: process.env.PASSWORD
    }
  }
}
```

Souvikns:
**Description**

This PR is updating the `glee.config.js` file to support configuration for multiple protocols.

- users can pass mqtt authentication parameters now from `glee.config.js` file instead of env variables. 
- users can now pass HTTP authentication token from `glee.config.js` file wich will be helpful for #319 

**Related issue(s)**

Fixes #347

### DIFF
--- a/docs/config-file.md
+++ b/docs/config-file.md
@@ -12,7 +12,7 @@ export default async function () {
 }
 ```
 
-This function must return an object with protocol specific configuration which would look like this: 
+This function must return an object with the following shape:
 
 
 ```js
@@ -26,7 +26,7 @@ export default async function () {
 
 ```
 
-Here is an example of the `glee.config.js` file for refence: 
+Here is an example of a `glee.config.js` file for reference: 
 ```js
 export default async function () {
   return {
@@ -38,10 +38,7 @@ export default async function () {
       },
       client: {
         authentication: {
-          http: () => {
-            const token = process.env.TOKEN
-            return token
-          }
+          token: () => process.token
         }
       }
     },
@@ -65,16 +62,21 @@ export default async function () {
 ```
 Every protocol has different configuration needs so each protocol has unique configurations:
 
-### Websockets
+### Websocket Server
 |Field|Description|
 |--|--|
-|websocket.server|Websocket server specific configuration|
-|websocet.client| Websocket client specific configuration|
-|wesocket.server.adapter| The Glee adapter to use for the WebSocket server. Defaults to a "native" WebSocket implementation. Other allowed values are `socket.io` (to use the [Socket.IO](https://socket.io/) Glee adapter) or a reference to a custom adapter.|
+|websocket.server|Websocket server-specific configuration|
+|websocket.server.adapter| The Glee adapter to use for the WebSocket server. Defaults to a "native" WebSocket implementation. Other allowed values are `socket.io` (to use the [Socket.IO](https://socket.io/) Glee adapter) or a reference to a custom adapter.|
 |websocket.server.httpServer|  A custom HTTP server of your own. E.g., an [Express](https://expressjs.com/en/4x/api.html) server or any object that implements the [http.Server](https://nodejs.org/api/http.html#http_class_http_server) interface.   |
-|webocket.server.port| The port to use when binding the WebSocket server. This is useful when your server is behind a proxy and the port exposed for consumption is not the same as the port your application should be bound to. Defaults to the port specified in the selected AsyncAPI server.|
-|websocket.client.authentication.http| The http token to send for authentication of glee application.
+|websocket.server.port| The port to use when binding the WebSocket server. This is useful when your server is behind a proxy and the port exposed for consumption is not the same as the port your application should be bound to. Defaults to the port specified in the selected AsyncAPI server.|
 
+
+### Websocket Client
+
+|Field|Description|
+|----|------|
+|websocket.client|Websocket client-specific configuration|
+|websocket.client.authentication.token| The http token to send for authentication of glee application.|
 
 ### Cluster 
 |Field|Description|
@@ -87,9 +89,9 @@ Every protocol has different configuration needs so each protocol has unique con
 
 |Field|Description|
 |---|---|
-|mqtt.authentication| mqtt authentication configuratation|
+|mqtt.authentication| MQTT authentication configuration|
 |mqtt.authentication.cert| Client certificate
-|mqtt.authentication.clientId| mqtt client Id for authentication
-|mqtt.authentication.userPassword| username and password paramters for authentication|
+|mqtt.authentication.clientId| MQTT client Id for authentication
+|mqtt.authentication.userPassword| username and password parameters for authentication|
 |mqtt.authentication.userPassword.username| username parameter
 |mqtt.authentication.userPassword.password| password parameter
--- a/docs/config-file.md
+++ b/docs/config-file.md
@@ -12,30 +12,84 @@ export default async function () {
 }
 ```
 
-This function must return an object with the following shape:
+This function must return an object with protocol specific configuration which would look like this: 
+
 
 ```js
 export default async function () {
   return {
-    websocket: {
-      httpServer: customServer, // A custom HTTP server of your own.
-      adapter: 'native', // Default. Can also be 'socket.io' or a reference to a custom adapter.
-      port: process.env.PORT,
-    },
-    cluster: {
-      adapter: 'redis',
-      name: 'cluster', // Default. Name of your cluster.
-      url: 'redis://localhost:6379', // Server URL used by adapter for clustering
-    }
+    websocket: {},
+    mqtt: {},
+    cluster: {}
   }
 }
+
 ```
+
+Here is an example of the `glee.config.js` file to refer to. 
+```js
+export default async function () {
+  return {
+    websocket: {
+      server: {
+        httpServer: customServer, // A custom HTTP server of your own.
+        adapter: "native", // Default. Can also be 'socket.io' or a reference to a custom adapter.
+        port: process.env.PORT,
+      },
+      client: {
+        authentication: {
+          http: () => {
+            const token = process.env.TOKEN
+            return token
+          }
+        }
+      }
+    },
+    cluster: {
+      adapter: "redis",
+      name: "cluster", // Default. Name of your cluster.
+      url: "redis://localhost:6379", // Server URL used by adapter for clustering
+    },
+    mqtt: {
+      authentication: {
+        cert: () => fs.readFileSync('./cert')
+        clientId: '123',
+        userPassword: {
+          username: 'user1',
+          password: 'pass12'
+        }
+      }
+    }
+  };
+}
+```
+Every protocol has different configuration needs so each protocol has unique configurations:
+
+### Websockets
+|Field|Description|
+|--|--|
+|websocket.server|Websocket server specific configuration|
+|websocet.client| Websocket client specific configuration|
+|wesocket.server.adapter| The Glee adapter to use for the WebSocket server. Defaults to a "native" WebSocket implementation. Other allowed values are `socket.io` (to use the [Socket.IO](https://socket.io/) Glee adapter) or a reference to a custom adapter.|
+|websocket.server.httpServer|  A custom HTTP server of your own. E.g., an [Express](https://expressjs.com/en/4x/api.html) server or any object that implements the [http.Server](https://nodejs.org/api/http.html#http_class_http_server) interface.   |
+|webocket.server.port| The port to use when binding the WebSocket server. This is useful when your server is behind a proxy and the port exposed for consumption is not the same as the port your application should be bound to. Defaults to the port specified in the selected AsyncAPI server.|
+|websocket.client.authentication.http| The http token to send for authentication of glee application.
+
+
+### Cluster 
+|Field|Description|
+|--|--|
+|cluster.adapter| The Glee cluster adapter to use for communication between instances. Defaults to Redis Pub/Sub ("redis"). Can be a reference to a custom adapter.|
+|cluster.name|The name of the cluster. Defaults to "cluster".|
+|cluster.url|The url of the server to be used by the adapter. In case of "redis" adapter, it's the url of the Redis server.|
+
+### MQTT
 
 |Field|Description|
 |---|---|
-|websocket.adapter|The Glee adapter to use for the WebSocket server. Defaults to a "native" WebSocket implementation. Other allowed values are `socket.io` (to use the [Socket.IO](https://socket.io/) Glee adapter) or a reference to a custom adapter.
-|websocket.httpServer|A custom HTTP server of your own. E.g., an [Express](https://expressjs.com/en/4x/api.html) server or any object that implements the [http.Server](https://nodejs.org/api/http.html#http_class_http_server) interface.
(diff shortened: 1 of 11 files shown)


src/lib/index.d.ts
```diff
@@ -1,48 +1,68 @@
...
-    httpServer?: any,
-    adapter?: WebSocketServerType | typeof GleeAdapter,
-    port?: number,
-  },
-  cluster?: GleeClusterAdapterConfig
+export type MqttAdapterConfig = {
+  authentication?: {
+    cert?: string
+    userPassword?: {username: string, password: string}
+    clientId?: string
```
Souvikns, line 31: I added three different parameters for authentication because I think MQTT allows all three forms of authentication and whatever is specified by the user will be used.

src/adapters/socket.io/index.ts
```diff
@@ -18,10 +18,11 @@ class SocketIOAdapter extends Adapter {
   }
 
   _connect(): Promise<this> {
-    return new Promise((resolve) => {
+    return new Promise(async (resolve) => {
```
Souvikns, line 21:
```suggestion
    return new Promise( (resolve) => {
```

I think promise executor function should not be `async`, how do we call an async function in a promise executor function?

src/adapters/mqtt/index.ts
```diff
@@ -1,124 +1,175 @@
...
+              })
+            })
+          }
+
+          resolve(this)
         })
       })
+    }
+
+    return connectClient()
```
Souvikns, line 147: I am trying to promisify this function, but getting this error -> 
```
Type 'unknown' is not assignable to type 'this'.
  'this' could be instantiated with an arbitrary type which could be unrelated to 'unknown'.
```

I am most probably missing something. @fmvilas can you help me on this.

Souvikns: Solved it, was missing the return type.

Souvikns:
Need your review @fmvilas

docs/config-file.md
```diff
@@ -12,30 +12,84 @@ export default async function () {
 }
 ```
 
-This function must return an object with the following shape:
+This function must return an object with protocol specific configuration which would look like this:
```
fmvilas, line 15: I think we should leave it as it was. There's also the undocumented possibility to add the following configuration keys:

* GLEE_DIR
* GLEE_LIFECYCLE_DIR
* GLEE_FUNCTIONS_DIR
* ASYNCAPI_FILE_PATH

See https://github.com/asyncapi/glee/blob/master/src/lib/configs.ts#L41.

Also `cluster` is not a protocol but the configuration of the cluster mode.

docs/config-file.md
```diff
@@ -12,30 +12,84 @@ export default async function () {
...
+  return {
+    websocket: {},
+    mqtt: {},
+    cluster: {}
+  }
+}
+
+```
+
+Here is an example of the `glee.config.js` file for refence:
```
fmvilas, line 29:
```suggestion
Here is an example of a `glee.config.js` file for reference: 
```

docs/config-file.md
```diff
@@ -12,30 +12,84 @@ export default async function () {
...
+        httpServer: customServer, // A custom HTTP server of your own.
+        adapter: "native", // Default. Can also be 'socket.io' or a reference to a custom adapter.
+        port: process.env.PORT,
+      },
+      client: {
+        authentication: {
+          http: () => {
+            const token = process.env.TOKEN
+            return token
+          }
```
fmvilas, line 44:
```suggestion
          http: () => process.env.TOKEN
```

docs/config-file.md
```diff
@@ -12,30 +12,84 @@ export default async function () {
...
+  };
 }
 ```
+Every protocol has different configuration needs so each protocol has unique configurations:
+
+### Websockets
+|Field|Description|
+|--|--|
+|websocket.server|Websocket server specific configuration|
+|websocet.client| Websocket client specific configuration|
```
fmvilas, line 72:
```suggestion
|websocket.client| Websocket client-specific configuration|
```

Can you also please move it below to group it with the other `websocket.client` configs? It's hard to read as it is.

docs/config-file.md
```diff
@@ -12,30 +12,84 @@ export default async function () {
...
+|wesocket.server.adapter| The Glee adapter to use for the WebSocket server. Defaults to a "native" WebSocket implementation. Other allowed values are `socket.io` (to use the [Socket.IO](https://socket.io/) Glee adapter) or a reference to a custom adapter.|
+|websocket.server.httpServer|  A custom HTTP server of your own. E.g., an [Express](https://expressjs.com/en/4x/api.html) server or any object that implements the [http.Server](https://nodejs.org/api/http.html#http_class_http_server) interface.   |
+|webocket.server.port| The port to use when binding the WebSocket server. This is useful when your server is behind a proxy and the port exposed for consumption is not the same as the port your application should be bound to. Defaults to the port specified in the selected AsyncAPI server.|
+|websocket.client.authentication.http| The http token to send for authentication of glee application.
```
fmvilas, line 76: Shouldn't this allow for other types of authentication too? I sometimes use a bearer token, and other times I use an API token/key. Depending on the server, it may require me to pass a token like `Authentication: bearer [TOKEN]`, `Authentication: token [TOKEN]`, `Authentication: api-key [token]`, etc. I think being able to say how we authenticate (using a header or a query param) and customize the header value is the minimum required here.

docs/config-file.md
fmvilas: BTW, [WebSocket authentication is always HTTP](https://www.rfc-editor.org/rfc/rfc6455#section-4) so it makes more sense to use `token` instead of `http`.

docs/config-file.md
```diff
@@ -12,30 +12,84 @@ export default async function () {
...
+Every protocol has different configuration needs so each protocol has unique configurations:
+
+### Websockets
+|Field|Description|
+|--|--|
+|websocket.server|Websocket server specific configuration|
+|websocet.client| Websocket client specific configuration|
+|wesocket.server.adapter| The Glee adapter to use for the WebSocket server. Defaults to a "native" WebSocket implementation. Other allowed values are `socket.io` (to use the [Socket.IO](https://socket.io/) Glee adapter) or a reference to a custom adapter.|
+|websocket.server.httpServer|  A custom HTTP server of your own. E.g., an [Express](https://expressjs.com/en/4x/api.html) server or any object that implements the [http.Server](https://nodejs.org/api/http.html#http_class_http_server) interface.   |
+|webocket.server.port| The port to use when binding the WebSocket server. This is useful when your server is behind a proxy and the port exposed for consumption is not the same as the port your application should be bound to. Defaults to the port specified in the selected AsyncAPI server.|
```
fmvilas, line 75:
```suggestion
|websocket.server.port| The port to use when binding the WebSocket server. This is useful when your server is behind a proxy and the port exposed for consumption is not the same as the port your application should be bound to. Defaults to the port specified in the selected AsyncAPI server.|
```

docs/config-file.md
```diff
@@ -12,30 +12,84 @@ export default async function () {
...
 }
 ```
+Every protocol has different configuration needs so each protocol has unique configurations:
+
+### Websockets
+|Field|Description|
+|--|--|
+|websocket.server|Websocket server specific configuration|
+|websocet.client| Websocket client specific configuration|
+|wesocket.server.adapter| The Glee adapter to use for the WebSocket server. Defaults to a "native" WebSocket implementation. Other allowed values are `socket.io` (to use the [Socket.IO](https://socket.io/) Glee adapter) or a reference to a custom adapter.|
```
fmvilas, line 73:
```suggestion
|websocket.server.adapter| The Glee adapter to use for the WebSocket server. Defaults to a "native" WebSocket implementation. Other allowed values are `socket.io` (to use the [Socket.IO](https://socket.io/) Glee adapter) or a reference to a custom adapter.|
```

docs/config-file.md
```diff
@@ -12,30 +12,84 @@ export default async function () {
...
-|websocket.httpServer|A custom HTTP server of your own. E.g., an [Express](https://expressjs.com/en/4x/api.html) server or any object that implements the [http.Server](https://nodejs.org/api/http.html#http_class_http_server) interface.
-|websocket.port|The port to use when binding the WebSocket server. This is useful when your server is behind a proxy and the port exposed for consumption is not the same as the port your application should be bound to. Defaults to the port specified in the selected AsyncAPI server.
-|cluster.adapter|The Glee cluster adapter to use for communication between instances. Defaults to Redis Pub/Sub ("redis"). Can be a reference to a custom adapter.
-|cluster.name|The name of the cluster. Defaults to "cluster".
-|cluster.url|The url of the server to be used by the adapter. In case of "redis" adapter, it's the url of the Redis server.
\ No newline at end of file
+|mqtt.authentication| mqtt authentication configuratation|
```
fmvilas, line 90:
```suggestion
|mqtt.authentication| MQTT authentication configuration|
```

fmvilas (changes requested):

docs/config-file.md
Souvikns: I am keeping on one kind of authentication for this PR, since this PR just focuses on updating the config file, In #319 I will add support for all the authentication methods supported for websocket.

Souvikns: Have removed the `websocket.client` documentation for now.

docs/config-file.md
Souvikns: I have removed the client example as well, and will be tackling it in #319

src/adapters/mqtt/index.ts
```diff
@@ -20,105 +19,157 @@ class MqttAdapter extends Adapter {
...
+            serverBinding.lastWill.message
+            ? serverBinding.lastWill.message
+            : undefined,
+        retain:
+          serverBinding &&
+            serverBinding.lastWill &&
+            serverBinding.lastWill.retain
+            ? serverBinding.lastWill.retain
+            : undefined,
+      },
```
KhudaDad414, line 83:
```suggestion
      clientId: serverBinding?.clientId ?? mqttOptions?.authentication?.clientId,
      clean: serverBinding?.cleanSession,
      will: serverBinding?.will && {
        topic: serverBinding?.lastWill?.topic,
        qos: serverBinding?.lastWill?.qos,
        payload: serverBinding?.lastWill?.message,
        retain: serverBinding?.lastWill?.retain,
      },
```
@Souvikns I can't see why we are not using optional chaining here 🤔

docs/config-file.md
```diff
@@ -14,28 +14,71 @@ export default async function () {
...
-      adapter: 'redis',
-      name: 'cluster', // Default. Name of your cluster.
-      url: 'redis://localhost:6379', // Server URL used by adapter for clustering
+      adapter: "redis",
+      name: "cluster", // Default. Name of your cluster.
+      url: "redis://localhost:6379", // Server URL used by adapter for clustering
+    },
+    mqtt: {
+      authentication: {
+        cert: () => fs.readFileSync('./cert')
```
fmvilas, line 47: I think this function should be `async`:

```suggestion
        cert: async () => fs.readFileSync('./cert')
```

examples/dummy/glee.config.js
```diff
@@ -11,5 +12,10 @@ export default async function () {
     //   name: 'gleeCluster',
     //   url: 'redis://localhost:6379'
     // }
+    mqtt: {
+      authentication: {
+        cert: () => fs.readFileSync('./mosquitto.org.crt')
```
fmvilas, line 17:
```suggestion
        cert: async () => fs.readFileSync('./mosquitto.org.crt')
```

src/adapters/mqtt/index.ts
```diff
@@ -20,105 +19,157 @@ class MqttAdapter extends Adapter {
   }
 
   async connect(): Promise<this> {
-    return this._connect()
+    return await this._connect()
```
fmvilas, line 22: You don't have to `await` it:

```suggestion
    return this._connect()
```

src/adapters/ws/index.ts
```diff
@@ -12,121 +12,121 @@ class WebSocketsAdapter extends Adapter {
   }
 
   async connect(): Promise<this> {
-    return this._connect()
+    return await this._connect()
```
fmvilas, line 15:
```suggestion
    return this._connect()
```

src/lib/adapter.ts
```diff
@@ -148,6 +149,23 @@ class GleeAdapter extends EventEmitter {
     return this._serverUrlExpanded
   }
 
+  async resolveConfig(protocol: string) {
```
fmvilas, line 152: Would it make sense to call it `resolveProtocolConfig` instead?

```suggestion
  async resolveProtocolConfig(protocol: string) {
```

src/lib/adapter.ts
```diff
@@ -148,6 +149,23 @@ class GleeAdapter extends EventEmitter {
...
+    if(!this.glee.options[protocol]) return undefined
+    const protocolConfig = {...this.glee.options[protocol]}
```
fmvilas, line 154: Why are you making a copy of the object? 🤔

fmvilas (changes requested):

src/lib/adapter.ts
Souvikns: I am making a copy of the original config object so that I still have the unresolved functions. Any adapter can access it by doing `this.glee.options`. I did this because If any adapter wants to resolve the config functions every time a message is received, it can do so. (I don't know if this case will ever happen)

fmvilas: Oh, alright 👍 Sounds good!

Souvikns:
@fmvilas I have made all the changes that was requested, let me know if we are good to merge.

fmvilas (approved):
LGTM 👍 🚀

KhudaDad414 (approved):
LGTM!

PR #363, opened 2022-11-28
fix: make some improvements

KhudaDad414:
**Description**
I was working on an example project to get myself familiarize with glee I think these changes will fix a few bugs and make some improvements.
- glee will support multiple messages in channels.
- the glee types will be available for typescript projects.


src/adapters/ws/index.ts
```diff
@@ -93,15 +93,17 @@ class WebSocketsAdapter extends Adapter {
...
+            
+            ws.on('message', (payload) => {
+              const msg = this._createMessage(pathname, payload)
+              this.emit('message', msg, ws)
+            })
+
+            this.emit('server:connection:open', { name: this.name(), adapter: this, connection: ws, channel: pathname, request })
           })
 
           this.emit('connect', { name: this.name(), adapter: this, connection: ws, channel: pathname })
```
fmvilas, line 109: We should get rid of this `connect` event. As per documentation, `connect` is only triggered when Glee is a client but in this case it behaves as a server.

src/index.ts
```diff
@@ -67,7 +67,7 @@ export default async function GleeAppInitializer () {
     if (channel.hasPublish()) {
       const operationId = channel.publish().json('operationId')
       if (operationId) {
-        const schema = channel.publish().message().payload().json()
+        const schema = {oneOf: channel.publish().messages().map(message => message.payload().json())} as any;
```
fmvilas, line 70:
```suggestion
        const schema = {oneOf: channel.publish().messages().map(message => message.payload().json())} as any
```

src/index.ts
```diff
@@ -78,7 +78,7 @@ export default async function GleeAppInitializer () {
       }
     }
     if (channel.hasSubscribe()) {
-      const schema = channel.subscribe().message().payload().json()
+      const schema = {oneOf: channel.subscribe().messages().map(message => message.payload().json())} as any;
```
fmvilas, line 81:
```suggestion
      const schema = {oneOf: channel.subscribe().messages().map(message => message.payload().json())} as any
```

src/lib/configs.ts
```diff
@@ -44,9 +44,7 @@ async function loadConfigsFromFile() {
     ASYNCAPI_FILE_PATH = projectConfigs.ASYNCAPI_FILE_PATH ? path.resolve(GLEE_DIR, projectConfigs.ASYNCAPI_FILE_PATH) : ASYNCAPI_FILE_PATH
     return projectConfigs
   } catch (e) {
-    if (e.code !== 'ERR_MODULE_NOT_FOUND') {
       return console.error(e)
```
fmvilas, line 47:
```suggestion
    return console.error(e)
```

src/lib/lifecycleEvents.ts
```diff
@@ -79,7 +79,7 @@ export async function run(lifecycleEvent: string, params: GleeFunctionEvent) {
     const responses = await Promise.all(handlers.map(info => info.fn(params)))
     
     responses.forEach(res => {
-      if (res.send) {
+      if (res?.send) {
         res.send.forEach((event: GleeFunctionReturnSend) => {
```
fmvilas, line 83: Wouldn't it make more sense to do this directly?

```suggestion
        res?.send?.forEach((event: GleeFunctionReturnSend) => {
```

This way we remove the `if` clause.

fmvilas (changes requested):

fmvilas (approved):
LGTM 👍 

Pinging @Souvikns @sudoshreyansh @Ruchip16 @atharvagadkari05 too in case they want to review.

Souvikns (approved):

PR #367, opened 2022-12-12
feat: support multiple file extentions

### LINKED ISSUE #23: Support different asyncapi file names: asyncapi.json and asyncapi.yml

KhudaDad414:
**Description**
- support `.yaml`, `.json` and `.yml` file extensions for the AsyncAPI file.
- add core configuration to the config file.

fixes #23


KhudaDad414:
@Souvikns need your feedback here. 😁

Souvikns (dismissed):

src/lib/configs.ts
```diff
@@ -50,6 +50,26 @@ async function loadConfigsFromFile() {
...
+      if(indexOfFoundFile === -1){
+      indexOfFoundFile = index
+      } else {
+        throw new Error('Multiple AsyncAPI files found at the root of your project. Please make sure to choose one in your config file.')
+      }
+    }
+  }
+  
+  if(indexOfFoundFile === -1){
+    throw new Error('AsyncAPI file was not found at the root of your project. you can set its path in config file.')
```
fmvilas, line 69: Would be great to provide help on how "you can set its path in config file". Have a look at https://wix-ux.com/when-life-gives-you-lemons-write-better-error-messages-46c5223e1a2f.

Also, when this error throws, it will be showing the stack trace and all the stuff that comes with throwing an error. This information is not relevant as it's not a failure in the code. Please consider "catching" the error in a function above and simply show a beautiful and informative error message using https://github.com/asyncapi/glee/blob/master/src/lib/logger.ts#L123.

src/lib/configs.ts
```diff
@@ -50,6 +50,26 @@ async function loadConfigsFromFile() {
...
+  }
+  return path.resolve(baseDir, files[indexOfFoundFile])
```
fmvilas, line 71: This is a way to avoid the "Function Call Object Injection Sink" warning:

```suggestion
  return path.resolve(baseDir, files.at(indexOfFoundFile))
```

fmvilas (changes requested):
You've got additional warnings and I added a few comments on how you can solve them.

fmvilas (changes requested):

KhudaDad414:
@fmvilas thanks for the review. 🙇  I noticed some configurations weren't documented. I have added them and also changed the format so it matches the config file's overall format. anything you want to add?

src/lib/configs.ts
```diff
@@ -38,16 +44,30 @@ async function loadConfigsFromFile() {
...
+    ASYNCAPI_FILE_PATH = projectConfigs.glee?.asyncapiFilePath ?? ASYNCAPI_FILE_PATH
     return projectConfigs
   } catch (e) {
     return console.error(e)
   }
 }
 
+export function findSpecFile(baseDir: string): string{
+  const files = ['asyncapi.yaml', 'asyncapi.json', 'asyncapi.yml']
+  const foundFiles = files.filter(file => existsSync(path.resolve(baseDir, file)))
```
fmvilas, line 59: You should not be using `existsSync` because:

1. It only tests the existence of the path but it doesn't test if you have read permission. So it may happen that the file is there but you can't open it.
2. It tests the existence of the **path**. A path might be a file or a directory. We want to make sure it's a file.

Therefore I recommend using [fs.accessSync](https://nodejs.org/api/fs.html#fsaccesssyncpath-mode) and [fs.statSync](https://nodejs.org/api/fs.html#fsstatsyncpath-options) (which allows you to check for [isFile()](https://nodejs.org/api/fs.html#statsisfile)).

fmvilas (changes requested):
I found another problem. Other than that, seems ready 👍

src/lib/configs.ts
```diff
@@ -13,41 +14,68 @@ let GLEE_FUNCTIONS_DIR: string
...
+    // No error logging is required since we expect accessSync to fail most of the time.
+    return false
+  }
+}
 /**
  * Loads the configuration from glee project.
  */
 async function loadConfigsFromFile() {
-  if (!existsSync(GLEE_CONFIG_FILE_PATH)) return 
+  if (!isFileReadable(GLEE_CONFIG_FILE_PATH)) return
```
fmvilas, line 49: Good catch 💯

src/lib/configs.ts
```diff
@@ -13,41 +14,68 @@ let GLEE_FUNCTIONS_DIR: string
...
+export function findSpecFile(baseDir: string): string{
+  const files = ['asyncapi.yaml', 'asyncapi.json', 'asyncapi.yml']
+  const foundFiles = files.filter(file => isFileReadable(path.resolve(baseDir, file)))
+  
+  if (foundFiles.length === 1) {
+    return path.resolve(baseDir, foundFiles[0])
+  }  else if(foundFiles.length > 1) {
+    errorMessage = "Multiple AsyncAPI files found. Please choose one in you config file (https://github.com/asyncapi/glee/blob/master/docs/config-file.md)."
+  } else {
+    errorMessage = "Unable fo find the AsyncAPI file. Please make sure it's in your project's directory or set its path in the config file(https://github.com/asyncapi/glee/blob/master/docs/config-file.md)."
```
fmvilas, line 74:
```suggestion
    errorMessage = "Unable fo find the AsyncAPI file. Please make sure it's in your project's directory or set its path in the config file (https://github.com/asyncapi/glee/blob/master/docs/config-file.md)."
```

fmvilas (changes requested):
Last minor change. It's done already 🚀

fmvilas (approved):
Forgot to approve 🤦

PR #370, opened 2022-12-19
fix: functions may not return

fmvilas:
**Description**

Functions may or may not return. We have to accept functions that return undefined or do not return.


fmvilas:
@Souvikns @KhudaDad414 please have a look 🙏

jonaslagoni (approved):

PR #372, opened 2023-01-10
feat: support HTTP/SaaS function trigger

### LINKED ISSUE #25: Add support for an HTTP/FaaS runtime

KhudaDad414:
**Description**
- Adds `invoke` alongside `reply` and `send`. `invoke` can trigger HTTP/HTTPS endpoints. It enables Glee to support FaaS functions.
-  `invoke` can also be configured by setting the `operationId` to the URL of the FaaS function and customising it with an `x-glee-invoke` extension.
- Adds a prettier config file so we can enforce the formatting rules.
- Moves some misplaced `dependencies` to `devDependencies`.
- Create a prettier config file so we can enforce some styling rules that are not being enforced by es-lint.

**Related issue(s)**
Fixes #25


package.json
```diff
@@ -41,6 +41,7 @@
     "@types/qs": "^6.9.7",
     "ajv": "^6.12.6",
     "async": "^3.2.0",
+    "axios": "^1.2.2",
```
fmvilas, line 44: I'd use https://www.npmjs.com/package/node-fetch or similar so when Node.js introduces stable support for `fetch` the code will not be different.

fmvilas:
@KhudaDad414 did you close it accidentally or?

KhudaDad414:
I reset the branch since I am taking a different solution and I guess GitHub closed it automatically. 😆 I will reopen this soon.

src/lib/functions.ts
```diff
@@ -82,58 +52,54 @@ export async function trigger({
...
-        channel: msg.channel,
+      res?.reply?.forEach((msg) => {
+        message.reply({
+          payload: msg.payload,
+          headers: msg.headers,
+          channel: msg.channel
+        })
+      })
+      res?.invoke?.forEach((request) => {
+        httFetch(request, handleResponse)
```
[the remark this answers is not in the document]
KhudaDad414: In my defence, this PR is still in draft, so... 🤣

KhudaDad414:
@Souvikns would love to have your opinion here. 🙇

Souvikns (approved):
Looks good, really like how you are generating functions from the spec to use existing `triggerFunction` logic

KhudaDad414:
@fmvilas a review would be appreciated. 🙇

docs/functions.md
```diff
@@ -37,6 +37,7 @@ export default async function (event) {
 |---|---|---|
 |send|array&lt;[OutboundMessage](#anatomy-of-an-outbound-message)&gt;|A list of outbound messages to send when the processing of the inbound event has finished. All clients subscribed to the given channel/topic will receive the message.
 |reply|array&lt;[OutboundMessage](#anatomy-of-an-outbound-message)&gt;|A list of outbound messages to send as a reply when the processing of the inbound event has finished. This is useful when the target of your message is the sender of the inbound event. Note, however, that this only works when you're running Glee as a server. For example, using `reply` when receiving a WebSocket message is fine and the reply will exclusively go to the client that sent the message. However, if you're receiving a message from an MQTT broker, `reply` will work exactly the same way as `send` above, and will send the message to all the clients subscribed to the given channel/topic.
+|invoke|array&lt;[InvokeRequest](#anatomy-of-an-invoke-message)&gt;|A list of requests for reaching an HTTP/HTTPS endpoint. This is useful when you have a software as a service (SaaS) function and want it to be treated as one of Glee's functions.
```
fmvilas, line 40:
```suggestion
|invoke|array&lt;[InvokeRequest](#anatomy-of-an-invoke-message)&gt;|A list of requests for reaching an HTTP/HTTPS endpoint. This is useful when you have a Serverless function and want it to be treated as one of Glee's functions.
```

docs/functions.md
```diff
@@ -48,6 +49,17 @@ export default async function (event) {
...
 |server|string|The server/broker you want to send the message to. Defaults to `event.serverName`, i.e., the same server as the received event.
 
+##### Anatomy of an invoke message
+
+|Attribute|Type|Description|
+|---|---|---|
+|url|string|The URL of your HTTP/SaaS end-point.
+|headers|object&lt;string,string&gt;|The headers/metadata of the message you want to send.
+|method|string|The HTTP method that you want your endpoint to be called with.
+|ignoreResponse|boolean|By default, Glee assumes that the response is in JSON format and adheres to the [Functions](#functions) return type structure. If you do not wish for Glee to process the response, you can set the this option to `false`.
```
fmvilas, line 59:
```suggestion
|ignoreResponse|boolean|By default, Glee assumes that the response is in JSON format and adheres to the [Functions](#functions) return type structure. If you do not wish for Glee to process the response, you can set this option to `false`.
```

docs/functions.md
```diff
@@ -48,6 +49,17 @@ export default async function (event) {
...
+
+> NOTE: Glee uses [got](https://github.com/sindresorhus/got) under the hood. Therefore, you have the ability to personalize your request with any options that [got](https://github.com/sindresorhus/got#documentation) supports.
```
fmvilas, line 61: How? 🤔

docs/functions.md
```diff
@@ -61,4 +73,23 @@ channels:
       ...
 ```
 
-Glee maps the `onHello` operation to the `functions/onHello.js` file.
\ No newline at end of file
+Glee maps the `onHello` operation to the `functions/onHello.js` file.
+
+## Using FaaS functions as glee functions
```
fmvilas, line 78:
```suggestion
## Using Serverless functions as Glee functions
```

docs/functions.md
```diff
@@ -61,4 +73,23 @@ channels:
...
+
+Glee can treat your FaaS function as its own by setting `operationId` to your function URL and customizing it with the `x-glee-invoke` extension:
```
fmvilas, line 80:
```suggestion
Glee can treat your Serverless function as its own by setting `operationId` to your function URL and customizing it with the `x-glee-invoke` extension:
```

docs/functions.md
```diff
@@ -61,4 +73,23 @@ channels:
...
+```yaml
+channels:
+  hello:
+    publish:
+      operationId: https://domain.come/path/to/your/function
+      x-glee-invoke:
+        method: get
+        headers:
+          Content-Type: application/json
+        ignoreResponse: true # By default, Glee processes the response and raises an error if it does not conform to the format specified by [functions](#functions) return type.
```
fmvilas, line 91: It would be cool to note that this extension is not required. By default, we'll call the URL on `operationId` with the `POST` method, `Content-Type: application/json` header, and the message as the body.

Also, note that we can't send a request body when the method is `GET`. So we have to think what to do in this case. Do we serialize it as query params along with the URL or do we completely avoid sending the message in the body?

src/index.ts
```diff
@@ -43,12 +46,13 @@ export default async function GleeAppInitializer () {
     showFunctionsDir: GLEE_FUNCTIONS_DIR !== resolve(GLEE_DIR, 'functions'),
   })
 
-  await registerFunctions(GLEE_FUNCTIONS_DIR)
-  await registerLifecycleEvents(GLEE_LIFECYCLE_DIR)
-
   const parsedAsyncAPI = await getParsedAsyncAPI()
   const channelNames = parsedAsyncAPI.channelNames()
 
+  await generateFunctions(parsedAsyncAPI, GLEE_DIR)
```
fmvilas, line 52: Why did you choose to generate "empty" functions from the AsyncAPI file instead of interpreting the values directly? 🤔

fmvilas (changes requested):

docs/functions.md
KhudaDad414: Got it. I guess serialising the message makes sense here. but URLs are limited to 2048 characters which I guess can create untraceable errors (working fine for some requests and failing when the length exceeds 2048).
I think we need to respect the REST API rules/conventions here and ignore the message in GET requests.

src/index.ts
KhudaDad414: I don't get it. 🤔

docs/functions.md
fmvilas: Yeah, I would do this too. Honestly, I think most people will use POST. If not all.

src/index.ts
fmvilas: I was expecting that during the [Function `register` process](https://github.com/asyncapi/glee/blob/master/src/lib/functions.ts#LL54), you would read the AsyncAPI document searching for this kind of remote functions (those with URL) and would register them there. Instead, you're creating empty functions with an `invoke` property: https://github.com/asyncapi/glee/pull/372/files#diff-6f611215e3d2978b5127e7c36198526b54c97440b27b9fb25742c9a7ebfa0d0bR144.

I'm not saying one is better than the other, just curious why you took that route.

KhudaDad414: I would say registering them in the `register` function would be much faster since there is no I/O involved, plus there is no conversion between URLs and file names. 🤦 
let me try that...

docs/functions.md
KhudaDad414: clarified in the new commits.

KhudaDad414:
@fmvilas suggested changes have been applied. a 👁️ would be 🙏.

KhudaDad414:
@Souvikns @fmvilas another round of review 🙇

KhudaDad414:
@fmvilas conflicts have been resolved.

KhudaDad414:
closing this for two reasons:
1) Since all of the remote servers should be included in the AsyncAPI file, If somebody wants this feature they can already do this by including the SaaS/HTTP endpoint as an `http` server and glee will handle it with the `http` client adapter.

2) Glee should support the functionalities that are provided by the `spec` first then we can think about side cases.

PR #373, opened 2023-01-11
fix: server required when send

### LINKED ISSUE #371: Error when the server is not specified
https://github.com/asyncapi/glee/pull/272 had the side effect of specifying `server` in `send` messages mandatory.
I mean this worked with previous versions:
```
{
  send: [{
    payload: 'my message'
  }]
}
```
but now it gives:
```
x Cannot read property 'protocol' of null
    at file:///Users/khudadad/Projects/asyncapi/glee/dist/lib/functions.js:107:73
    at Array.forEach (<anonymous>)
    at file:///Users/khudadad/Projects/asyncapi/glee/dist/lib/functions.js:105:111
    at Generator.next (<anonymous>)
    at fulfilled (file:///Users/khudadad/Projects/asyncapi/glee/dist/lib/functions.js:4:58)
```

and I have to specify the server even if there is only one server:
```
{
   send: [{
      server: 'websockets',
      payload: 'my message'
  }]
}
```
If this was intentional, we need to update the docs and show better error messages.
I think we need to implement some kind of fallback to the default server if there is only one server. 🤔 

cc: @Souvikns

Souvikns:
**Description**
Server parameter should be specified when glee functions return a send object. This PR is changing the `OutboundMessageSchema` so that the `server` property is required. 

**Related issue(s)**

Fixes #371


Souvikns:
@KhudaDad414 let me know your thoughts

KhudaDad414 (approved):
LGTM! IMO, it would be great to not make it required if there is only one server. but on the other hand it just makes sense to mention the server in case of `send`. so no strong feeling, one way or another 😆

docs/functions.md
```diff
@@ -46,7 +46,7 @@ export default async function (event) {
 |payload|string|The payload/body of the message you want to send.
 |headers|object&lt;string,string&gt;|The headers/metadata of the message you want to send.
 |channel|string|The channel/topic you want to send the message to. Defaults to `event.channel`, i.e., the same channel as the received event.
-|server|string|The server/broker you want to send the message to. Defaults to `event.serverName`, i.e., the same server as the received event.
+|server|string|The server/broker you want to send the message to. Must be present when function returns a `send` event.
```
fmvilas, line 49: Why can't we default it to the same event as the one used to receive the event? 🤔

Souvikns: > Why can't we default it to the same event as the one used to receive the event? 🤔

We are doing this in the case of the Reply event, according to docs both reply and send events have the same type which is an `array<OutboundMessage>`. In the Reply event, the message is sent to the same channel that it came from. but send allows you to send it to 
- the same channel 
- and some other channel or channels (all channels in a server)

maybe we do not need a reply event, we can just return an array of events that could have a server property or not, if it does not have one we reply and if it does we send it to that specific server.

fmvilas: But server != connection. `reply` uses the same connection to reply. For instance, in the case of WS or HTTP, it replies to the specific client who made the "request", not all the clients connected to that server, which is what `send` would do.

In any case, what we're changing here is that we're removing the default value. I honestly don't see a reason to do it. For instance, I receive a message from an MQTT broker and, as a result, I want to send another one to the broker. I'd use `send` there because even though `reply` would do the same it might give the impression that it's going to respond to the message, which is not really possible on MQTT.

src/lib/functions.ts
```diff
@@ -109,13 +108,17 @@ export async function trigger({
...
+      if (msg.server) {
+        const serverProtocol = parsedAsyncAPI.server(msg.server).protocol().toLowerCase()
+        isBroadcast = localServerProtocols.includes(serverProtocol) && !isRemoteServer(parsedAsyncAPI, msg.server)
+      }
       app.send(new GleeMessage({
         payload: msg.payload,
         headers: msg.headers,
         channel: msg.channel || message.channel,
         serverName: msg.server,
+        connection: message.connection,
```
Souvikns, line 121: If no server name is passed then the message is sent to the connection of the incoming message.

src/lib/functions.ts
```diff
@@ -109,13 +108,17 @@ export async function trigger({
 
     res?.send?.forEach((msg) => {
       const localServerProtocols = ['ws', 'wss', 'http', 'https']
-      const serverProtocol = parsedAsyncAPI.server(msg.server).protocol().toLowerCase()
-      const isBroadcast = localServerProtocols.includes(serverProtocol) && !isRemoteServer(parsedAsyncAPI, msg.server)
+      let isBroadcast: boolean = true
```
fmvilas, line 111: You don't need the type there:

```suggestion
      let isBroadcast = true
```

fmvilas (changes requested):

src/lib/functions.ts
fmvilas: But this is always assigning the connection to the `message.connection`. Shouldn't we assign it ONLY when `msg.server` is not defined?

fmvilas (changes requested):

fmvilas:
I think we're not moving forward with his PR, right? Maybe we should close it? I'm saying it because of this discussion: https://github.com/asyncapi/glee/pull/373#discussion_r1069625185.

Souvikns:
Currently, when we send we need to specify the server, can we agree that it is the correct behavior as we should be passing in the server name when we use send operation?

fmvilas:
> Currently, when we send we need to specify the server, can we agree that it is the correct behavior as we should be passing in the server name when we use send operation?

Do we need to? I thought it was defaulting to the server where the message was received 🤔

Souvikns:
Yeah, I think we are already doing that, currently, this gives an error if the server is not specified, so in this PR I moved it to an if condition if the server is not specified we will not run this, and `broadcast` will be false. 

https://github.com/asyncapi/glee/blob/bc0d45f805442c013e6a3a294c533155564d5274/src/lib/functions.ts#L115

fmvilas:
🤔 I'm confused. We can get the server from `message.serverName` if `msg.server` is not present. That's one way to overcome this issue:

```ts
 const serverProtocol = parsedAsyncAPI.server(msg.server || message.serverName).protocol().toLowerCase() 
```

Isn't that enough?

Souvikns:
> 🤔 I'm confused. We can get the server from `message.serverName` if `msg.server` is not present. That's one way to overcome this issue:
> 
> ```ts
>  const serverProtocol = parsedAsyncAPI.server(msg.server || message.serverName).protocol().toLowerCase() 
> ```
> 
> Isn't that enough?


Yeah, we can do that too, I really didn't think about it 😄. I have made the changes I think we can merge this now.

fmvilas (changes requested):
I think we can simplify the code a bit more :)

PR #374, opened 2023-01-16
feat: update config to support multiple servers

Souvikns:
**Description**

This PR is trying to update the config file to support configs for multiple servers, the config file would look something like this. 

```ts
export default async function(){
  return {
    ws: {
      auth: ({serverName}) => {
        const c = {
          websockets: {
            token: process.env.TOKEN
          }
        }
        return c[serverName]
      }
    },
    mqtt: {
      auth: {
        username: process.env.USERNAME,
        password: process.env.USERNAME
      }
    }
  }
}
```

all auth function in any protocol, could be a object or a function, if it is a function it will  receive serverName and parsedAsyncAPI parameter so that the user can add custom logic for passing auth config


Souvikns:
Since `glee.config.js` file is a javascript file we care actually make the pass function as config objects, so we have have two possible way - 

#### authentication function returns config 
```ts
import fs from 'fs'

export default async function() {
    return {
        authentication: (servername, channelname, parsedAsyncAPI, config) => {
            if(servername === 'ws-client') {
                return {
                    token: process.env.TOKEN
                }
            }

            if(servername === 'mqtt') {
                return {
                    cert: fs.readFileSync('./cert')
                }
            }
        }
    }
}

```
Here we call the authentication function on every adapter and pass in different parameters, it up to the users to return necessary config according to the server name.

#### pass in authentication config object that has helper functions to assist the users 
```ts
import configFile from 'configfile.json'

export default async function() {
  return {
    authentication: (config) => {
      configFile.forEach(c => config.addConfig(c.servername, c.config))

      // another way 
      config.addConfig('server-name', {
        token: process.env.TOKEN
      })
    }
  }
}
``` 
Here we can pass in an `AuthenticationConfig` class with helper functions to help the user pass in config values with server names as keys.

docs/config-file.md
```diff
@@ -62,19 +59,25 @@ export default async function () {
...
-          password: 'pass12'
+      auth: ({serverName, parsedAsyncAPI}) => {
+        if (serverName === 'mqtt') {
+          return {
+            cert: async () => fs.readFileSync('./cert')
+            clientId: '123',
+            userPassword: {
+              username: 'user1',
+              password: 'pass12'
+            }
```
fmvilas, line 70: I think this could be easily flattened:

```suggestion
            username: 'user1',
            password: 'pass12'
```

docs/config-file.md
```diff
@@ -119,21 +121,51 @@ These configurations apply to Glee itself, rather than any specific protocol.
...
-|mqtt.authentication.clientId| MQTT client Id for authentication
-|mqtt.authentication.userPassword| username and password parameters for authentication|
-|mqtt.authentication.userPassword.username| username parameter
-|mqtt.authentication.userPassword.password| password parameter
+|mqtt.auth| MQTT authentication configuration|
+|mqtt.auth.cert| Client certificate
+|mqtt.auth.clientId| MQTT client Id for authentication
+|mqtt.auth.userPassword| username and password parameters for authentication|
+|mqtt.auth.userPassword.username| username parameter
+|mqtt.auth.userPassword.password| password parameter
```
fmvilas, line 129: As mentioned above, this could be easily flattened.

docs/config-file.md
```diff
@@ -62,19 +59,25 @@ export default async function () {
...
-      authentication: {
-        cert: async () => fs.readFileSync('./cert')
-        clientId: '123',
-        userPassword: {
-          username: 'user1',
-          password: 'pass12'
+      auth: ({serverName, parsedAsyncAPI}) => {
+        if (serverName === 'mqtt') {
+          return {
+            cert: async () => fs.readFileSync('./cert')
```
fmvilas, line 65:
```suggestion
            cert: async () => fs.readFileSync('./cert', 'utf-8')
```

src/adapters/kafka/index.ts
```diff
@@ -93,6 +94,20 @@ class KafkaAdapter extends Adapter {
       },
     })
   }
+
+  private async resolveAuthConfig() {
+    const config = this.glee.options?.kafka
+    if (!config) return undefined
```
fmvilas, line 100:
```suggestion
    if (!config) return
```

src/adapters/mqtt/index.ts
```diff
@@ -64,12 +66,12 @@ class MqttAdapter extends Adapter {
       },
       keepalive: serverBinding?.keepAlive,
       username: userAndPasswordSecurityReq
-        ? mqttOptions?.authentication?.userPassword.username
+        ? mqttOptions?.userPassword.username
```
fmvilas, line 69: Regarding flattening:

```suggestion
        ? mqttOptions?.username
```

fmvilas (changes requested):
Left some suggestions

src/adapters/mqtt/index.ts
```diff
@@ -168,6 +170,17 @@ class MqttAdapter extends Adapter {
     })
   }
 
+  private async resolveAuthConfig() {
```
KhudaDad414, line 173: The `resolveAuthConfig()` function logic is repeated in all of the adapters. I feel like we can make it a part of the `Adapter` class.

KhudaDad414 (commented):
Looks great!

src/adapters/mqtt/index.ts
Souvikns: Different protocols have different config types, I tried writing it in adapters logic but it was a lot of nested if statements. I think it will be confusing when support for new protocols are added. But if we implement it on the protocol-specific adapters it would be easier as you don't need to know about other protocol config types.

KhudaDad414: I see. then maybe it's time to make a few changes to the `configs.ts` file itself. I think it should be a single truth source for all config-related things. we can abstract all of the logic there with a method like `getServerConfig(serverName: string)`. 🤔   WDYT?

Souvikns: are you suggesting that we move the config to the server name instead of protocol name so the config file would look something like this 

```ts
  return {
    'server-name': {
      protocol: 'ws'
      // rest of the config. 
    }
  }
}

```

should we move this to an issue and explore the possibilities.

KhudaDad414: I think the current format of the config file makes sense. All I am saying is that we should abstract the config parsing logic for different protocols and servers to https://github.com/asyncapi/glee/blob/master/src/lib/configs.ts

Souvikns: Now I got it, sure I will try this and let you know.

Souvikns: @KhudaDad414 what about now, I have moved the parsing logic to the adapter class.

KhudaDad414 (dismissed):
LGTM!

docs/config-file.md
```diff
@@ -119,21 +119,50 @@ These configurations apply to Glee itself, rather than any specific protocol.
...
+|kafka.auth.cert| Client certificate
+|kafka.auth.clientId| Kafka client Id for authentication
+|kafka.auth.rejectUnauthorized | Boolean flag for accepting the valid SSL certificates
+|kafka.auth.username| The username to use during authentication.
+|kafka.auth.password| The password to use during authentication.
+
+
+### Auth Config
+
+Most client like `ws`,`kafka` and `mqtt` has auth config which is used for passing auth parameters, all these config could be an object or an function that returns the specific object defined by each protocol.
```
fmvilas, line 143:
```suggestion
Most clients like `ws`,`kafka`, and `mqtt` have `auth` fields that are used for passing auth parameters. All these configurations can be an object or a function that returns the specific object defined by each protocol.
```

fmvilas: Suggestion: go to https://grammarly.com and create a free account. Then install the browser extension and the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=znck.grammarly). It will help you write better :)

docs/config-file.md
```diff
@@ -119,21 +119,50 @@ These configurations apply to Glee itself, rather than any specific protocol.
...
+
+
+
+```js
+/** Both of these examples are correct **/
+
```
fmvilas, line 149: I don't think that's needed:

```suggestion
```

examples/crypto-websockets/client/db.json
```diff
@@ -1 +1 @@
-[{"time":1663935946665,"price":130,"status":"started"},{"time":1663935947684,"price":140,"status":"intransit"},{"time":1663935948690,"price":130,"status":"intransit"},{"time":1663935949698,"price":140,"status":"intransit"},{"time":1663935950710,"price":150,"status":"intransit"},{"time":1663935951716,"price":130,"status":"intransit"},{"time":1663935952724,"price":100,"status":"intransit"},{"time":1663935953737,"price":110,"status":"intransit"},{"time":1663935954742,"price":140,"status":"intransit"}]
\ No newline at end of file
+[{"time":1675676576262,"price":60,"status":"started"},{"time":1675676577270,"price":90,"status":"intransit"},{"time":1675676578272,"price":120,"status":"intransit"},{"time":1675676579280,"price":80,"status":"intransit"},{"time":1675676580285,"price":110,"status":"intransit"},{"time":1675676581291,"price":130,"status":"intransit"},{"time":1675676582295,"price":170,"status":"intransit"},{"time":1675676583299,"price":150,"status":"intransit"},{"time":1675676584308,"price":130,"status":"intransit"}]
```
fmvilas, line 1: You shouldn't be adding this file :)

fmvilas (changes requested):
Left some suggestions

fmvilas (approved):

Issue #377, opened 2023-01-17
Authentication layer for glee, support for different authentication frameworks 


Souvikns:
Currently, as a user when we are using glee to build a WebSocket server the only way of authenticating is by writing your own authentication logic in the functions. Which is repetitive and not a good solution when you have a large application, with the addition of WebSocket and HTTP adapters, glee needs support for authentication and authorization. Right now glee needs some kind of Authentication middleware for both server and client adapters. 

Currently, client adapters are able to send authentication parameters with the help of [glee config](https://github.com/asyncapi/glee/blob/master/docs/config-file.md), and this needs to be updated so that it can support multiple authentication and authorization process. 

https://github.com/asyncapi/glee/blob/16c99b0d62fb4ac7bd2348330f7eb347756698c1/examples/dummy/glee.config.js#L15-L19

For starters, we need to support some popular authorization processes-

- **Token-based authentication**: In this method, the client requests an access token from the server, which can then be used to access protected resources. The token is sent with each websocket request and is used by the server to verify the client's identity and determine their authorization.
- **Basic authentication**: This method uses the same basic authentication process as the HTTP protocol, where the client sends an HTTP request with a header that contains a username and password. The server verifies the credentials and returns a response with the requested resource if the user is authenticated.
- **Certificate-based authentication**: In this method, the client and server use digital certificates to establish trust and verify identity. The client presents a certificate to the server, which then verifies the certificate's validity and authenticity.
- **Session-based authentication**: This method works similarly to session-based authentication in the HTTP protocol, where the server creates a session for the user and associates it with an identifier. The client then sends the identifier with each websocket request, allowing the server to identify the user and determine their authorization.
- **OAuth 2.0 authentication**: OAuth 2.0 is an open standard for authorization that provides secure access to resources by delegating access to third-party applications. The client requests access to a resource from the server, and the server returns an authorization code, which the client can then use to obtain an access token.

Souvikns:
Just putting a thought out there, since glee is using middlewares we could use https://www.npmjs.com/package/passport to implement authentication.

octonawish-akcodes:
I am interested in this under gsoc 2023, can you guide me to get started with this issue.

oviecodes:
@Souvikns @fmvilas @KhudaDad414 I would love to give this a try, I believe it could pass as a first issue for me. let me know what you think. Thanks

Infamia2334:
[...]

PR #439, opened 2023-05-18
feat: glee authentication

oviecodes:
**Description**
Glee Authentication PR
In order to introduce authentication to Glee, I've been looking at the middleware, which offers the most opportunity as we advance.

One idea is to create an authentication middleware that reads the user’s authentication.
The authentication file will be similar to a lifecycle event, it will have the `server`and the `channel` export along with a default export.

In the `authMiddleware.ts` file:

```
    // authMiddleware.ts
    import { AsyncAPIDocument } from "@asyncapi/parser";
    import { MiddlewareCallback } from "./index.js";
    import GleeMessage from "../lib/message.js";
    export default (asyncapi: AsyncAPIDocument) =>
      (event: GleeMessage, next: MiddlewareCallback) => {
        const componentNames = asyncapi.components();
        console.log("security schemes", componentNames.securitySchemes());
        if (Object.keys(componentNames.securitySchemes()).length == 0)
          return next();
        if (componentNames.securitySchemes()) {
          console.log("Authentication should happen here", event.channel);
          //check if channel matches, then
          //call user authentication function from file
          return next();
        }
        return next();
      }; 
```

The authentication file will take the following format

```
    export default async function ({ glee, connection }) {
      //user authentication logic
    }
    
    export const servers = ['websockets']
    export const channels = ['user/signedup']
```

`In the index.ts file`:

```   
 //index.ts
    ...
    import authMiddleware from "./middlewares/authMiddleware.js";
    ...
    app.use(authMiddleware(parsedAsyncAPI));
    app.useInbound(authMiddleware(parsedAsyncAPI));
```

`app.useInbound()` doesn’t exist in the codebase so it’ll be created

We’ll create a file that allows us to parse the user authentication similar to what Glee does with lifecycle events.

Another idea I’ve been trying to explore is to see if Glee middleware is extendable, if it’s possible we could allow the user to write a middleware then read such middleware and pass it to Glee during initialization.

```
    export default (asyncapi) =>
      (event, next) => {
        //authentication logic
          return next();
        }
        return next();
      };
```

Since  `app.useOutbound()` already supports channel property as follows:

    app.useOutBound("/", authMiddleware(parsedAsyncAPI))

Restriction of auths to a particular channel should be easy. We could make `app.useInbound()` exactly like `app.useOutbound()` but for incoming messages.

**Related issue(s)**
`Discusses #377 `

@Souvikns  @KhudaDad414  @fmvilas 
#377


oviecodes:
No problems @KhudaDad414  the questions are necessary.

1. I was going for authentication on each message, but now you mentioned it Authentication on connection makes the most sense. I was thinking about authenticating different endpoints hence the middleware idea.

2. In the middleware system, handling the authentication method depends on the user of glee, all they have to do is write the authentication logic in a file. However, when using the adapters, the responsibility falls on glee.

I will take a look at the WebSockets adapters compared to others. I think an authentication property could be added there. Further inputs are welcomed @Souvikns @KhudaDad414 I will really appreciate it

Issue #436, opened 2023-05-18
Glee is crashing when I pass custom server from config file. 


Souvikns:
#### Describe the bug
I am trying to use expressjs with glee, to see how well it would work and for that I need to pass in a custom HTTP server with an express app object, when I pass in the custom server glee crashed and throws this error 

```
⚡ Server websockets is ready to accept connections.
TypeError: Cannot convert undefined or null to object
    at hasOwnProperty (<anonymous>)
    at getOrSetAsyncId (node:internal/async_hooks:423:7)
    at Object.connectionListener [as connection] (node:_http_server:596:5)
    at file:///Users/postman/Documents/Learning/glee-custom-server/node_modules/@asyncapi/glee/dist/lib/util.js:121:60
    at Generator.next (<anonymous>)
    at file:///Users/postman/Documents/Learning/glee-custom-server/node_modules/@asyncapi/glee/dist/lib/util.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (file:///Users/postman/Documents/Learning/glee-custom-server/node_modules/@asyncapi/glee/dist/lib/util.js:3:12)
    at resolveFunctions (file:///Users/postman/Documents/Learning/glee-custom-server/node_modules/@asyncapi/glee/dist/lib/util.js:115:45)
    at file:///Users/postman/Documents/Learning/glee-custom-server/node_modules/@asyncapi/glee/dist/lib/util.js:118:13
    at Generator.next (<anonymous>)
    at file:///Users/postman/Documents/Learning/glee-custom-server/node_modules/@asyncapi/glee/dist/lib/util.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (file:///Users/postman/Documents/Learning/glee-custom-server/node_modules/@asyncapi/glee/dist/lib/util.js:3:12)
    at resolveFunctions (file:///Users/postman/Documents/Learning/glee-custom-server/node_modules/@asyncapi/glee/dist/lib/util.js:115:45)
    at file:///Users/postman/Documents/Learning/glee-custom-server/node_modules/@asyncapi/glee/dist/lib/util.js:118:13

```

#### How to Reproduce
Create a new glee application using asyncapi cli `asyncapi new glee` and then create a `glee.config.js` file and write this code 

```js
const express = require('express')
const http = require('http')
const app = express()
const httpServer = http.createServer()

httpServer.listen(8000)

module.exports = () => {
    return {
        ws: {
          server: {
            httpServer: httpServer,
[...]

PR #445, opened 2023-05-29
feat: update config management

### LINKED ISSUE #392: Better config management
#### Reason/Context
When I want to add something to the config file currently I have to take the following steps:
1) update the `lib/index.d.ts` and add my config types.
2) update the `docs/config-file.md` to reflect the new configs that I have added/changed.
3) If no config is provided in my config file I have to hardcode the default values in adapters themselves.

When working with the codebase, it's a little hard to go back and forth between these files and make sure the are synced and the parsing logic is correct. 

#### Description
IMO we can make some improvements to the codebase and make it easier for the new contributors to understand the codebase by having a config class that is responsible for all three of those above.

1) The class should be documented correctly with `jsdoc` so we can generate the `doc/config-file.md` from it instead of maintaining it separately.

2) all of the parsing logic should be handled by this class itself. I shouldn't have to look around for the port of my server in three different places and finally hardcode the default value in each `Adapter`. 😄 

3) Providing default values for each configuration.

Souvikns:
**Description**
This PR aims to refactor how the config is handled and parsed in glee, no changes to how the config works or any changes to the config file itself, just refactoring the codebase for ease of use. 

**Related issue(s)**

Fixes #392

### DIFF
--- a/src/lib/config/index.ts
+++ b/src/lib/config/index.ts
@@ -1,0 +1,39 @@
+import ConfigDocs from "./docs.js"
+import ConfigGee from './glee.js'
+import ConfigMqtt from './mqtt.js'
+import ConfigWs from './ws.js'
+
+export class Config {
+    protected async resolve(ob) {
+        if (typeof ob !== 'function') {
+            return ob
+        }
+        return await ob()
+    }
+}
+
+export default class GleeConfig {
+    private _config: any
+    
+    constructor(config: any) {
+        this._config = config
+    }
+
+    get docs() {
+        return new ConfigDocs(this._config.docs)
+    }
+
+    get glee() {
+        return new ConfigGee(this._config.glee)
+    }
+
+    get mqtt() {
+        return new ConfigMqtt(this._config.mqtt)
+    }
+
+    get ws() {
+        return new ConfigWs(this._config.ws)
+    }
+}
+
+const config = new GleeConfig({})
--- a/src/lib/configs.ts
+++ b/src/lib/configs.ts
@@ -105,28 +105,3 @@ export function getConfigs(): { [key: string]: string } {
 }
 
 
-export class GlobalGleeConfig {
-  private _config: GleeConfig
-
-  constructor(config: GleeConfig) {
-    this._config = config
-  }
-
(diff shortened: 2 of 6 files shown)


src/lib/configs.ts
```diff
@@ -102,4 +102,31 @@ export function getConfigs(): { [key: string]: string } {
...
+
+  async init(protocol: string){
+
+    const defaultValues = {
+      docs: {
+        enabled: true,
+        folder: 'docs',
+        template: '@asyncapi/markdown-template'
+      }
+    }
```
Souvikns, line 123: We can define the default values like this in the init function where we can basically return default values for the config or the passed values.

src/lib/configs.ts
```diff
@@ -102,4 +102,31 @@ export function getConfigs(): { [key: string]: string } {
...
+
+    return {
+      ...this._config[protocol],
+      defaultValues[protocol]
+    }
+  }
```
Souvikns, line 129: I don't think we can do this, as this is currently giving me errors and like this we have to traverse the whole object to find what values to return according to the default or passed, this creates the same problem as it tries to parse the entire object, this has the same issue as https://github.com/asyncapi/glee/issues/436

Souvikns (commented):
I think we need to convert all the config objects to classes from interfaces, that way we can then define the default values, also currently I think only docs and glee config have default values, any other protocol config does not need default values.

What do you think @KhudaDad414

src/lib/config/index.ts
```diff
@@ -0,0 +1,39 @@
...
+        return new ConfigGee(this._config.glee)
+    }
+
+    get mqtt() {
+        return new ConfigMqtt(this._config.mqtt)
+    }
+
+    get ws() {
+        return new ConfigWs(this._config.ws)
+    }
```
Souvikns, line 36: This was a idea that I had, just writing it to show it once, what do you think @KhudaDad414

Souvikns:
@KhudaDad414 now that we are removing the client authentication from config I will be updating this accordingly, I think it will fix the issue we were having before and it will be a lot easier now.

PR #460, opened 2023-06-07
feat: glee authentication v1

oviecodes:
**Description**
This PR continues the discussion on the glee authentication issue #377 

We have moved on from the middleware approach proposed in PR #439 and have started exploring the possibilities of using `lifecycle events` for authentication. This PR details the progress that has been made using Lifecycle events.

Progress:
1. Creation of an entirely new lifecycle event (`onAuth`)
2. Successful emission of `auth` event  from websocket server

Challenges:
1. websocket proceeds to connect even when auth file has not finished running due to the nature of node.js events.

We're currently experimenting with potential fixes for this challenge.

**Related issue(s)**
`Discuses issue #377 `

@Souvikns @KhudaDad414


oviecodes:
The use of `lifecycle events` for authentication has been successful so far. The cornerstone behind this success is using events.

However, this creates a question of whether lifecycle methods should be the main method for authentication. There are a couple of other ways to carry out authentication.

1. Using a separate authentication directory to hold user authentication functions.
2. Using a config file e.g. a server `glee.config.js` file like the one available in the client. Although I believe this method will limit the user's ability to craft elaborate auth functions.
3. Lifecycle events which we already Implement in Glee.  But we will have another handler, rather than the `runLifecycleEvents` handler, something like `runAuth`, which will specifically respond to `onAuth` events.

I think the third method might be the best. But it's up for discussion.
@Souvikns @KhudaDad414 .

KhudaDad414:
@oviecodes I also think the third option is the best option. this method can be used to authenticate http and web sockets. just a few things:

1) can you update the PR with some clean code? it is really hard to understand the changes with all of that auto-formatting 😆  
2) one challenge that we have here is that we need a way to make it clear that this option only belongs to http and web-sockets. 🤔 the current implementation doesn't have any indication of that. passing an auth function in `glee.config.js` would solve this problem but I am not sure what kind of consequences that can have. can you explore that as well.

3) we should use the info that we get from the asyncapi files `security` object and utilize that to make it a little bit easier for user to authenticate. I think we can do more here rather than just giving them the `header`. 🤔

Souvikns:
>Lifecycle events which we already Implement in Glee. But we will have another handler, rather than the runLifecycleEvents handler, something like runAuth, which will specifically respond to onAuth events.

I agree as well that the third option is better, I just have one thing to point out which is what the user sees, Authentication is different for both a server and a client, the client just needs to send the required parameter for authentication and server needs to run a auth functions to actually check those parameters. going with the third option we have two separate ways for authentication for the server and client (for the user as it is two different ways to achieve it). So what I am saying is should we consider updating the current way of client authentication?

Souvikns (commented):
@oviecodes I think you are using some different code formatting tool and it is formatting every file, it is too hard to review could you revert those back or follow the convention that Glee is already following?

Souvikns:
### We should have a separate place for Authentication 

We could have a separate place for authentication, instead of lifecycle events and config. We can have both server auth and client auth in the same place. 


```ts
/** auth/serverName.ts **/

export function serverAuth({headers, done}) {
  // write logic for authentication 
  done()
}

export function clientAuth() {
  return {
    username: '',
    password: ''
  }
}

```

- With this, we can remove the auth from the config. 
- Server and client auth stays together and the user does not have to change multiple things. 
- Since this follows the file system like `functions` we can find out which server the authentication file belongs to and thus figure out if it is `http` or `websocket` and call the functions accordingly. 

> @KhudaDad414, @oviecodes  what do you think?

oviecodes:
> ### We should have a separate place for Authentication
> We could have a separate place for authentication, instead of lifecycle events and config. We can have both server auth and client auth in the same place.

> > @KhudaDad414, @oviecodes  what do you think?

@Souvikns  That's a good approach, we'll have to create a separate folder for `authentication` alone, since the logic in the auth file has grown past  being housed in `lifecycle` files.

KhudaDad414:
> We could have a separate place for authentication, instead of lifecycle events and config. We can have both server auth and client auth in the same place.


@Souvikns that's a great idea. it would solve lots of issues and makes a lot of sense.

> Should this change affect other adapters or should it be for just the WS and HTTP adapters?

@oviecodes For now lets focus on HTTP and WS. that's our goal for GSoC. we can extend it if the need arise.

KhudaDad414:
@oviecodes thanks to @Souvikns the formatting issue should be fixed with https://github.com/asyncapi/glee/pull/469 the formatting issue seems to be solved. let's move your changes to a new PR so we can discuss it further. 😀

PR #466, opened 2023-06-20
fix: issue with spectral and jest

Souvikns:
**Description**

So according to https://github.com/asyncapi/parser-js#testing-with-jest we need to add some `moduleMapper` rule to our `package.json` file since we are using parser.js and there is an issue with jest and spectral where jest is unable to understand package exports. 

This is causing these PRs to fail.

- https://github.com/asyncapi/glee/pull/465
- https://github.com/asyncapi/glee/pull/464
- https://github.com/asyncapi/glee/pull/463

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,15797 @@
 {
   "name": "@asyncapi/glee",
   "version": "0.21.4",
-  "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


KhudaDad414 (approved):
LGTM!

PR #474, opened 2023-07-10
feat: glee auth v2

### LINKED ISSUE #377: Authentication layer for glee, support for different authentication frameworks

oviecodes:
**Description**
This PR continues the discussion on the glee authentication issue #377 , it continues the discussion about glee authentication, from #460 .

Following the idea from @Souvikns  to use a separate directory for authentication. Each file in this directory can hold both server and client authentication logic.

````js
/** auth/serverName.ts **/

export function serverAuth({headers, done}) {
  // write logic for authentication 
  done()
}

export function clientAuth() {
  return {
    username: '',
    password: ''
  }
}
````

**Steps to test run authentication**
- In the glee root directory, `cd examples/anime-http/server`,  
- The server directory, contains an `auth` directory that holds a `trendingAnimeServer.ts` file. The file contains some logic you can edit how you like but it must export the same functions as the code above
- `npm run dev` to start the server in dev mode
- The same logic applies to the client directory, `cd examples/anime-http/server`
- Then look for `auth/trendingAnime.js` update accordingly then start the server.

**Related issue(s)**
`Resolves #377`

@Souvikns @KhudaDad414

### CODE REVIEW

File: src/lib/userAuth.ts [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+      })
+    )
+  } catch (e) {
+    console.error(e)
+  }
+}
+export async function triggerAuth(params: GleeFunctionEvent) {
+  const { serverName, callback } = params
+
+  console.log('running auth')
```
KhudaDad414, line 52: It would be great to use https://www.npmjs.com/package/debug to log some debugging information of the process.

examples/anime-http/server/auth/trendingAnimeServer.ts
```diff
@@ -0,0 +1,26 @@
...
+  await axios.get("https://jsonplaceholder.typicode.com/todos/1")
+
+  console.log("running in auth file in auth folder")
+
+  console.log("headers", headers["authentication"])
+
+  // console.log("network data", res.data)
+
+  // callback(false, 401, "Unauthorized");
+  done(true)
```
KhudaDad414, line 17: When auth fails, the server crashes, it should show a nice error message to console and send `401` respond.

examples/anime-http/server/asyncapi.yaml
```diff
@@ -53,3 +58,17 @@ components:
...
+      bearerFormat: JWT
+    userPass:
+      type: userPassword
+    apiKey:
+      type: httpApiKey
+      name: api_key
+      in: header
+    cert:
+      type: apiKey
+      in: user
```
KhudaDad414, line 74: We definitly documentation to explain how each of these work with glee and show some snippets of code on how to set it up and do the authentecation process for each one.

test/lib/adapter.test.ts
```diff
@@ -163,103 +233,90 @@ describe('adapter', () => {
...
+      adapter.emit('close', {
+        channels: ['fake/channel'],
+        connection: RAW_CONN,
       })
-      adapter.emit('close', { channels: ['fake/channel'], connection: RAW_CONN })
     })
   })
-  
+
   describe('on("reconnect")', () => {
```
KhudaDad414, line 267: Why are we removing this test?

test/lib/adapter.test.ts
```diff
@@ -163,103 +233,90 @@ describe('adapter', () => {
...
-      app.on('adapter:close', (ev) => {
-        expect(ev.serverName).toStrictEqual(TEST_SERVER_NAME)
-        expect(ev.server).toStrictEqual(TEST_SERVER)
-        expect(ev.connection).toBeInstanceOf(GleeConnection)
-        expect(ev.connection.AsyncAPIServer).toStrictEqual(TEST_SERVER)
-        expect(ev.connection.rawConnection).toStrictEqual(RAW_CONN)
-        expect(ev.connection.channels).toStrictEqual(['fake/channel'])
-        expect(ev.connection.serverName).toStrictEqual(TEST_SERVER_NAME)
-        expect(ev.connection.parsedAsyncAPI).toStrictEqual(TEST_ASYNCAPI_DOCUMENT)
+      const adapter = initializeAdapter()
```
KhudaDad414, line 259: and this one?

src/lib/userAuth.ts
```diff
@@ -0,0 +1,74 @@
+import { basename, extname, relative, join } from 'path'
```
KhudaDad414, line 1: not used imports

src/lib/userAuth.ts
```diff
@@ -0,0 +1,74 @@
...
+import { logWarningMessage } from './logger.js'
+import { GleeFunction, GleeFunctionEvent } from './index.js'
+import { pathToFileURL } from 'url'
+
+interface AuthFunctionInfo {
+  clientAuth?: GleeFunction
+  serverAuth?: GleeFunction
+}
+
+const { GLEE_DIR, GLEE_AUTH_DIR } = getConfigs()
```
KhudaDad414, line 14: not used imports

src/adapters/ws/server.ts
```diff
@@ -74,6 +74,52 @@ class WebSocketsAdapter extends Adapter {
     })
   }
 
+  private getSecurityReqs() {
```
KhudaDad414, line 77: extracting these steps into another class to handle auth related things for http and ws would be great.

KhudaDad414 (commented):
left some comments.

examples/anime-http/server/asyncapi.yaml
oviecodes: > We definitly documentation to explain how each of these work with glee and show some snippets of code on how to set it up and do the authentecation process for each one.

Could you explain further Khuda, it's kinda confusing @KhudaDad414

examples/anime-http/server/asyncapi.yaml
```diff
@@ -53,3 +58,17 @@ components:
         studio:
           type: string
           description: The studio of anime.
+  securitySchemes:
```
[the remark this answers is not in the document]
oviecodes: If there's an authentication function but there's no security in the asyncAPI file, then the server shouldn't emit the `auth` events, but I think we should add a warning just so they know that security isn't present. But if it's vice-versa, then authentication should fail outrightly, we throw an Error.

test/lib/adapter.test.ts
oviecodes: it's not removed, I refactored it. It was making use of the same logic so I moved that logic to a separate function because it was causing sonar to fail due to code duplication.

test/lib/adapter.test.ts
oviecodes: I just refactored it.

examples/anime-http/server/asyncapi.yaml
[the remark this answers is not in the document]
Souvikns: I think if there is no security schema then the auth function shouldn't even be loaded, and we throw an error if there is any security schema but no function.

examples/anime-http/server/auth/trendingAnimeServer.ts
```diff
@@ -0,0 +1,28 @@
...
+
+// //@ts-ignore
+// import { Message } from "@asyncapi/glee";
+import axios from "axios"
+
+export async function serverAuth({ headers, callback: done }) {
+  await axios.get("https://jsonplaceholder.typicode.com/todos/1", {
+    timeout: 5000,
+  })
+
```
Souvikns, line 11:
```suggestion
export async function serverAuth({ params, callback: done }) {
  await axios.get("https://jsonplaceholder.typicode.com/todos/1", {
    timeout: 5000,
  })
  
  params.getParams()
  // or
  params() it could be a function that just return the parameters that are needed. 

```

and we can parse the security requirement to just return what is needed.

examples/crypto-websockets/server/asyncapi.yaml
```diff
@@ -8,6 +8,12 @@ servers:
   websocket:
     url: ws://localhost:3000
     protocol: ws
+    security:
+      - tokens: []
+      - JWT: []
```
Souvikns, line 13: I might be wrong but shouldn't you reference the `securitySchemes` object.

oviecodes: No, that's not needed. The spec file matches them using the name. For each `security`, it checks if there's a matching `securityScheme`.

Souvikns (changes requested):
- Open the PR for review (I want to run the tests)
- There are some code smells as per `sonarcloud` resolve them 
- resolve the previous reviews that you have already fixed.

oviecodes:
The PR is failing, I'll resolve it once I'm back from college 😂

examples/anime-http/client/auth/trendingAnime.ts
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable no-undef */
+
+export async function clientAuth({ serverName }) {
+    console.log("serverName", serverName)
+  
+    return {
+      token: process.env.TOKEN,
+      oauth: process.env.OAUTH2,
```
Souvikns, line 8: Currently are just able to send `OAuth` token from the client side right, is glee able to handle OAuth on server side?

Souvikns (changes requested):
Everything looks good, I think you should now start working on the docs.

.eslintrc
```diff
@@ -43,7 +43,7 @@
         "max-nested-callbacks": "off",
         "sonarjs/no-identical-functions": "off",
         "@typescript-eslint/no-non-null-assertion": "off",
-        "@typescript-eslint/no-unused-vars": "off",
+        "@typescript-eslint/no-unused-vars": 1,
```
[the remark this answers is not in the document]
oviecodes: @Souvikns the issue i faced here has been existent on glee for a while, it doesn't recognize `@typescript-eslint/no-unused-vars` hence some variable that are not used in files such as `glee.ts`, `adapter.ts` and some other core files were giving sonar some issues

examples/anime-http/client/auth/trendingAnime.ts
oviecodes: @Souvikns could you help me understand the question? We have the `params.getOauth()` method to retrieve the oauth token, Is there something else that's needed?

KhudaDad414 (changes requested):
@oviecodes can you please add the documentation or a step by step guide on how can I get started with auth. from adding `security` field to asyncapi document to creating the files in glee. the file should be in `.md` folder in `docs` folder.

in a seperate PR though. 😆

KhudaDad414 (changes requested):
@oviecodes I can't run it on my local machine:
Screenshot 2023-08-10 at 16 33 32


### steps to reproduce.
1) checked out this branch.
2) removed the axios import and call from the server auth function.
3) run the anime-http server.
4) run
```bash
curl -X POST \
     -H "Content-Type: application/json" \
     -d '{
           "name": "Attack on Titan",
           "rating": "9.5",
           "genre": "Action",
           "studio": "WIT Studio"
         }' \
     http://localhost:8081/trendingAnime
```

KhudaDad414:
@oviecodes no I don't. but if I call `done(true)` in auth function it works fine. 🤔

oviecodes:
@KhudaDad414 you're suppose to evaluate the value gotten from the params, if it passes for authentication then you can go on to call done(true) if not you call done(false).

src/adapters/http/server.ts
```diff
@@ -30,14 +31,75 @@ class HttpAdapter extends Adapter {
...
       res.setHeader('Content-Type', 'application/json')
+
       const bodyBuffer = []
       let body: object
       req.on('data', (chunk) => {
         bodyBuffer.push(chunk)
       })
-      req.on('end', () => {
+
+      function done() {
```
KhudaDad414, line 43: please extract the authentication logic in another class. I am thinking something like 🤔 :
```ts
class Authenticator {
    AsyncAPIServer: any; 
    parsedAsyncAPI: any;
    serverName: string;
    constructor({...})

    authenticate(req: any): Promise<void>;
    checkAuthPresence(): boolean;
}

```

src/adapters/http/server.ts
```diff
@@ -30,14 +31,75 @@ class HttpAdapter extends Adapter {
...
+          done: (val: boolean, code?: number, message?: string) => {
+            if (val === true) {
+              resolveFunc(true)
+            } else if (val === false) {
+              rejectFunc({
+                code: code || 401,
+                message: message || 'Unauthorized',
+              })
+              return
+            }
```
KhudaDad414, line 60:
```suggestion
          done: (val: boolean, code: number = 401, message: string = 'Unauthorized') => {
            if (val) {
              resolveFunc(true)
            } else {
              rejectFunc({ code, message })
            }
```

KhudaDad414:
@oviecodes it should  be up to the user to authenticate. the only thing that we should care about is the value of `done()`.
In theory, I should be able to do this right?
```ts
export async function serverAuth({ authProps, callback: done }) {
  done(true)
}
```
or

```ts
export async function serverAuth({ authProps, callback: done }) {
  done(false)
}
```

KhudaDad414:
@oviecodes please open the docs PR when it is ready. I think it makes sense to review both of them together. I am having trouble reviewing this PR without the know hows. 😆

examples/anime-http/server/asyncapi.yaml
KhudaDad414: I meant we should have a documentation for the types of auths that can be done here. sorry didn't see your comment. 🤦

examples/crypto-websockets/client/auth/websockets.ts
```diff
@@ -0,0 +1,11 @@
+export async function clientAuth({ parsedAsyncAPI, serverName }) {
+    return {
```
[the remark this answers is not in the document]
oviecodes: yes @KhudaDad414

KhudaDad414 (approved):

Souvikns (approved):
Looks good 👍🏼, you can merge it when ready.

Souvikns:
@allcontributors please add @oviecodes for code

KhudaDad414:
@allcontributors please add @oviecodes for code

PR #485, opened 2023-08-01
feat: new config format and parser

### LINKED ISSUE #392: Better config management

Souvikns:
**Description**

Since we are moving the client auth from the `config file` to `auth` folder we need to update our config file format and update the parser as well to fetch these data differently. 

This PR is trying to achieve two things:

- [ ] New format for the config file (also the requirements for the config file now that auth parameters are removed)
- [ ] Update the config file parsing logic according the the new format 

**Related issue(s)**

Fixes #392

### DIFF
--- a/docs/config-file.md
+++ b/docs/config-file.md
@@ -19,7 +19,8 @@ export default async function () {
   return {
     glee: {},
     docs: {},
-    server: {}
+    server: {},
+    cluster: {}
   }
 }
 
@@ -45,6 +46,11 @@ export default async function () {
         adapter: 'native', // Default. Can also be 'socket.io' or a reference to a custom adapter.
         port: process.env.PORT,
       }
+    },
+    cluster: {
+      adapter: 'redis',
+      name: 'cluster', // Default. Name of your cluster.
+      url: 'redis://localhost:6379', // Server URL used by adapter for clustering
     }
   };
 }
--- a/docs/config-file.md
+++ b/docs/config-file.md
@@ -18,11 +18,8 @@ This function must return an object with the following shape:
 export default async function () {
   return {
     glee: {},
-    kafka: {},
-    websocket: {},
-    mqtt: {},
-    cluster: {},
-    http: {}
+    docs: {},
+    server: {}
   }
 }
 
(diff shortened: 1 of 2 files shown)


docs/config-file.md
```diff
@@ -18,11 +18,9 @@ This function must return an object with the following shape:
...
-    kafka: {},
-    websocket: {},
-    mqtt: {},
-    cluster: {},
-    http: {}
+    docs: {},
+    server: {},
+    cluster: {}
   }
 }
```
Souvikns, line 25: [This](https://github.com/asyncapi/glee/blob/b0c1670f777ad2a4dd0e33004d5ab78a72964463/docs/config-file.md) was the old format for `glee-config` and it was done to configure WebSocket servers for socket-io or pass in a custom server object. So building on that I am trying to add a new object named `server` where we would pass in the custom server config. The parser would take a server name from the spec and see if any config exists in the config file and then load that accordingly.

Souvikns (commented):
@KhudaDad414 I am trying to change the format for the `config-file` now that the `auth` parameters are being removed.

docs/config-file.md
KhudaDad414: It looks good, the only thing that worries me a little about this approach is writing an interface for it. I mean we want type suggestion/safety on the file right? with server name as a key I am not sure how we can write protocol specific interfaces. 🤔

Souvikns: I think we can still achieve that, I am not sure but I used notion API which had something like this kind of type 

```ts
interface T {
  server: {
    [name:string]: {
      httpAdapter: any,
      adapter: string,
      port: number
    }
  }
}

```

KhudaDad414: yeah, If type can be checked then let's do this. 👍

KhudaDad414:
@Souvikns what are we going to do with this? 🤔

Souvikns:
Since the config file has changed a lot, I am closing this and will open a new PR.

PR #517, opened 2023-09-08
fix: fix config resolution

### LINKED ISSUE #509: Specifying functionsDir or lifecyclesDir in config file results in an error
#### Describe the bug
An Error occurs when you try to specify the `functionsDir` or `lifecycleDir` in the `glee.config.ts` file as a result of glee not reading the files from the `.glee` directory whenever these values are specified.

#### How to Reproduce
In the client directory of any example in the Glee repo, create a `config.glee.ts` file, then specify the `functionsDir` as follows:

```ts
  return {
    glee: { // Glee core configurations
      lifecycleDir: './lifecycle',
      functionsDir: './functions', 
    }
  }
}
```

In the `functions` folder, create a function file with `.ts` extension.

You should get the following error:

Screenshot 2023-08-21 at 2 31 34 PM

#### Expected behavior
Glee should be able to resolve `.ts` files when `lifecycleDir` and `functionsDir` is specified in `glee.config.ts`

@Souvikns @KhudaDad414

oviecodes:
**Description**
This pull request will fix the configFile resolution issues, now `.ts` files can be used as in `glee.config.ts`. Glee will read from the `GLEE_DIR` for both `.ts` and `.js` config files as opposed to just reading from the `PROJECT_DIR`

**Related issue(s)**
fixes #509

@Souvikns @KhudaDad414


Souvikns (changes requested):
@oviecodes I think you should write a test for this, we already have a test for reading `asyncapi file` https://github.com/asyncapi/glee/blob/master/test/lib/config.spec.ts I think we should have one for reading the config as well.

oviecodes:
Alright, I think that should be a separate PR @Souvikns or should I include it here?

Souvikns:
I would suggest you include it in this PR.

oviecodes:
@Souvikns I've updated the tests, please check it out

Souvikns (approved):
Good Job @oviecodes 👍🏼

oviecodes:
Thanks @Souvikns can I merge this PR

PR #523, opened 2023-09-20
feat: update parser-js to v2

### LINKED ISSUE #409: upgrade parser to version 3(next-major)
#### Reason/Context
Following a recent discussion with @Souvikns and @fmvilas it seems like it's a plausible idea to use the latest version (v3.0.0-next-major.5) of @asynapi/parser-js in Glee.

#### Why?
* For solving #391 we have #407 which uses the latest version of modelina and without upgrading the parser in glee, we can't continue with this PR.
* Version 3 of the spec will be released soon and we need the new version of parser to support version 3 of the spec.
* The new intent-api implementation of parser-js in version 2 is way cooler than the current implementation :)

### Scope

- [x] Migrate to Parser v3
- [x] Update MQTT Adapter along with its examples.
- [ ] Update docs
- [ ] Update tests

Souvikns:
**Description**

This PR aims to upgrade parser-js to v2 to support spec 2.0

**Related issue(s)**

Fixes #409


src/adapters/http/client.ts
```diff
@@ -24,13 +24,13 @@ class HttpClientAdapter extends Adapter {
     const authConfig = await clientAuthConfig(this.serverName)
     const serverUrl = this.serverUrlExpanded
     for (const channelName of this.channelNames) {
-      const channelInfo = this.parsedAsyncAPI.channel(channelName)
-      const httpChannelBinding = channelInfo.binding('http')
-      const channelServers = channelInfo.servers()
+      const channelInfo = this.parsedAsyncAPI.channels().get(channelName)
```
fmvilas, line 27: Have you tried this? I see the `url` variable on line 34 is using the `channelName` but in v3 the channel name is not the address as it used to be in v2. I don't think this code is working 🤔

src/adapters/http/server.ts
```diff
@@ -93,7 +93,7 @@ class HttpAdapter extends Adapter {
         this.httpResponses.set(this.serverName, res)
         let { pathname } = new URL(req.url, serverUrl)
         pathname = pathname.startsWith('/') ? pathname.substring(1) : pathname
-        if (!this.parsedAsyncAPI.channel(pathname)) {
+        if (!this.parsedAsyncAPI.channels().get(pathname)) {
```
fmvilas, line 96: Same as above. The channel name is not necessarily the address. You should find the channel that has the `pathname` address.

src/adapters/http/server.ts
```diff
@@ -105,8 +105,8 @@ class HttpAdapter extends Adapter {
         const searchParams = { query }
         const payload = body
         const httpChannelBinding = this.parsedAsyncAPI
-          .channel(pathname)
-          .binding('http')
+          .channels().get(pathname)
```
fmvilas, line 108: Same as above.

src/adapters/mqtt/index.ts
```diff
@@ -127,7 +129,7 @@ class MqttAdapter extends Adapter {
 
   private subscribe(channels: string[]) {
     channels.forEach((channel) => {
-      const operation = this.parsedAsyncAPI.channel(channel).publish()
+      const operation = this.parsedAsyncAPI.channels().get(channel).json().subscribe()
```
fmvilas, line 132: Nope. Never access json properties like this. You're breaking the whole purpose of the new parser. Also, this won't work with v3 because there's no publish or subscribe in the new version. Instead, use the equivalent methods. Here, you should try to find the "send" operation that is associated with the channel `channel`. You have to iterate `this.parsedAsyncAPI.operations()`.

src/adapters/mqtt/index.ts
```diff
@@ -205,7 +207,7 @@ class MqttAdapter extends Adapter {
 
   _send(message: GleeMessage): Promise<void> {
     return new Promise((resolve, reject) => {
-      const operation = this.parsedAsyncAPI.channel(message.channel).subscribe()
+      const operation = this.parsedAsyncAPI.channels().get(message.channel).json().subscribe()
```
fmvilas, line 210: Same as above.

src/adapters/ws/client.ts
```diff
@@ -77,13 +77,12 @@ class WsClientAdapter extends Adapter {
   private getWsChannels() {
     const channels = []
     for (const channel of this.channelNames) {
-      if (this.parsedAsyncAPI.channel(channel).hasBinding('ws')) {
-        if (this.parsedAsyncAPI.channel(channel).hasServers()) {
+      if (this.parsedAsyncAPI.channels().get(channel).bindings().get('ws')) {
+        if (this.parsedAsyncAPI.channels().get(channel).servers().length !== 0) { // NOSONAR
```
fmvilas, line 81: Is this working for you? 🤔 

I think it should be:

```suggestion
        if (this.parsedAsyncAPI.channels().get(channel).servers().all().length !== 0) { // NOSONAR
```

src/adapters/ws/server.ts
```diff
@@ -232,8 +232,8 @@ class WebSocketsAdapter extends Adapter {
       )
 
       const wsChannelBinding = this.parsedAsyncAPI
-        .channel(pathname)
-        .binding('ws')
+        .channels().get(pathname)
```
fmvilas, line 235: Same as before. Pathname is not necessarily the name of the channel. It's the address for sure.

src/index.ts
```diff
@@ -59,7 +59,7 @@ export default async function GleeAppInitializer() {
   await registerAuth(GLEE_AUTH_DIR)
 
   const parsedAsyncAPI = await getParsedAsyncAPI()
-  const channelNames = parsedAsyncAPI.channelNames()
+  const channelNames = parsedAsyncAPI.channels().map(e => e.address())
```
fmvilas, line 62: I have mixed feelings about this 😅 I see what you're trying to do here but I think it's gonna bite you hard down the track because you're assuming that there can't be multiple channels with the same address, which is not true.

If I were you, I'd leave the array of `channelNames` as the array of channel names and would use it to retrieve the address whenever I need it.

fmvilas (changes requested):
Did a partial review as I have to leave but I think you have enough stuff to look on right now.

src/adapters/http/client.ts
Souvikns: In this PR I am not changing the logic to parse to v3, I am just updating the parser to v2, glee still only supports the spec v2 documents.

Souvikns: @fmvilas if you are talking about this 

```
        let url = `${serverUrl}/${channelName}`
```

I rechecked and this is giving me `ws://localhost:3000/price` as the output.

fmvilas: If you're testing [this document](https://github.com/asyncapi/glee/blob/master/examples/crypto-websockets/client/asyncapi.yaml#L19), this is going to work because in v3 channelName === address. They're the same thing. But in v3, channelName and address can be different (and most probably will be different). For instance:

```yaml
asyncapi: 3.0.0
...
channels:
  thePrice:
    address: '/price'
    ...
```

In this case, `channelName` would be equal to `thePrice` but the `address` would be `/price`. So you don't want to chain with `channelName` to form the URL but, instead, you want to chain with the `address`. Does that make sense?

test/middlewares/validate.test.ts
```diff
@@ -34,7 +38,7 @@ const TEST_ASYNCAPI_DOCUMENT = new AsyncAPIDocument({
   }
 })
 
-const schema = TEST_ASYNCAPI_DOCUMENT.channel('test/channel').publish().message().payload().json()
+const schema = TEST_ASYNCAPI_DOCUMENT.channels().get('test/channel')?.json().publish.message.payload
```
fmvilas, line 41:
```suggestion
const schema = TEST_ASYNCAPI_DOCUMENT.channels().get('test/channel').operations().filterByReceive()[0].messages().all()[0].payload()
```

src/adapters/mqtt/index.ts
```diff
@@ -42,22 +42,24 @@ class MqttAdapter extends Adapter {
   }
 
   private getSecurityReqs() {
-    const securityRequirements = (this.AsyncAPIServer.security() || []).map(
-      (sec) => {
-        const secName = Object.keys(sec.json())[0]
-        return this.parsedAsyncAPI.components().securityScheme(secName)
+
+    const parsedSecurityScehemes = this.parsedAsyncAPI.components().securitySchemes().all()
```
fmvilas, line 46:
```suggestion
    const parsedSecurityScehemes = this.AsyncAPIServer.securitySchemes().all()
```

src/adapters/mqtt/index.ts
```diff
@@ -42,22 +42,24 @@ class MqttAdapter extends Adapter {
...
+
+    let userAndPasswordSecurityReq
+    let X509SecurityReq
+
+    for (const security of parsedSecurityScehemes) {
```
fmvilas, line 51:
```suggestion
    for (const security of parsedSecuritySchemes) {
```

test/middlewares/validate.test.ts
Souvikns: I was running into an error in the validate function and now I realize is that in the docs it says that it returns `schema` [here](https://github.com/asyncapi/parser-api/blob/master/docs/api.md#message), but in the code it is returning `schemeInterface` [here](https://github.com/asyncapi/parser-js/blob/43a5b051b52ca87a519d66bb6c91307bf0f36556/src/models/v2/message.ts#L28) and tests are failing.

src/index.ts
Souvikns: I created a new property in the `Adapter class` called `channelAddresses` which returns the address and changed `channelNames` to return the `Id`'s  and when I need the URL I am doing something like `this.parsedAsyncAPI.channels().get(channelName).address()` this way we don't have to touch the logging logic for spec v3.

package.json
```diff
@@ -37,7 +37,7 @@
     "@asyncapi/generator": "^1.12.0",
     "@asyncapi/html-template": "^0.28.4",
     "@asyncapi/markdown-template": "^1.3.3",
-    "@asyncapi/parser": "^1.13.1",
+    "@asyncapi/parser": "^2.1.0",
```
KhudaDad414, line 40: lets use the latest version of parser. `v3.0.0-next-major-spec.3`

KhudaDad414 (commented):
Just one small change.

src/index.ts
```diff
@@ -59,7 +59,7 @@ export default async function GleeAppInitializer() {
   await registerAuth(GLEE_AUTH_DIR)
 
   const parsedAsyncAPI = await getParsedAsyncAPI()
-  const channelNames = parsedAsyncAPI.channelNames()
+  const channelNames = parsedAsyncAPI.channels().all().map(e => e.address())
```
fmvilas, line 62: You call it `channelNames` but you're getting the addresses. Can't you just get the channels names (ids) and then use it to get the address whenever you need the address?

src/index.ts
```diff
@@ -78,35 +78,34 @@ export default async function GleeAppInitializer() {
...
-            message: event,
+    const channel = parsedAsyncAPI.channels().get(channelName)
+    if (channel.operations().filterByReceive().length !==0) {
+
+      channel.operations().filterByReceive().forEach(operation => {
+        const receiveOperation = operation
+        const operationId = operation.operationId()
+        if (operationId) {
+          const schema = {
+            onOf: receiveOperation.messages().filterByReceive().map(m => m.payload())
```
fmvilas, line 89:
```suggestion
            oneOf: receiveOperation.messages().filterByReceive().map(m => m.payload())
```

src/index.ts
```diff
@@ -78,35 +78,34 @@ export default async function GleeAppInitializer() {
...
-          .subscribe()
-          .messages()
-          .map((message) => message.payload().json()),
-      } as any
-      app.useOutbound(channelName, validate(schema), json2string)
+    if (channel.operations().filterBySend().length !== 0) {
+      channel.operations().filterBySend().forEach(operation => {
+        const sendOperation = operation
+        const schema = {
+          onOf: sendOperation.messages().filterBySend().map(m => m.payload())
```
fmvilas, line 105:
```suggestion
          oneOf: sendOperation.messages().filterBySend().map(m => m.payload())
```

src/index.ts
```diff
@@ -78,35 +78,34 @@ export default async function GleeAppInitializer() {
...
-        app.use(channelName, validate(schema), (event, next) => {
-          triggerFunction({
-            app,
-            operationId,
-            message: event,
+    const channel = parsedAsyncAPI.channels().get(channelName)
+    if (channel.operations().filterByReceive().length !==0) {
+
+      channel.operations().filterByReceive().forEach(operation => {
+        const receiveOperation = operation
```
fmvilas, line 85: Do you really need this? I mean, do you get any benefit?

src/index.ts
```diff
@@ -78,35 +78,34 @@ export default async function GleeAppInitializer() {
...
-      const schema = {
-        oneOf: channel
-          .subscribe()
-          .messages()
-          .map((message) => message.payload().json()),
-      } as any
-      app.useOutbound(channelName, validate(schema), json2string)
+    if (channel.operations().filterBySend().length !== 0) {
+      channel.operations().filterBySend().forEach(operation => {
+        const sendOperation = operation
```
fmvilas, line 103: Same as above. Do you get any benefit out of this?

src/lib/adapter.ts
```diff
@@ -41,7 +42,7 @@ class GleeAdapter extends EventEmitter {
   constructor(
     glee: Glee,
     serverName: string,
-    server: Server,
+    server: Server | undefined,
```
fmvilas, line 45:
```suggestion
    server: Server?,
```

What about this?

fmvilas (changes requested):
Left some more comments.

src/adapters/http/client.ts
```diff
@@ -23,15 +23,15 @@ class HttpClientAdapter extends Adapter {
     let headers = {}
     const authConfig = await clientAuthConfig(this.serverName)
     const serverUrl = this.serverUrlExpanded
-    for (const channelName of this.channelNames) {
-      const channelInfo = this.parsedAsyncAPI.channel(channelName)
-      const httpChannelBinding = channelInfo.binding('http')
-      const channelServers = channelInfo.servers()
+    for (const channelName of this.channelAddresses) {
```
fmvilas, line 26:
```suggestion
    for (const channelName of this.channelNames) {
```

fmvilas (changes requested):
Added one more suggestion.

src/index.ts
```diff
@@ -78,36 +78,30 @@ export default async function GleeAppInitializer() {
...
-    if (channel.hasSubscribe()) {
+    })
+
+    channel.operations().filterBySend().forEach(operation => {
       const schema = {
-        oneOf: channel
-          .subscribe()
-          .messages()
-          .map((message) => message.payload().json()),
+        oneOf: operation.messages().filterBySend().map(m => m.payload())
```
KhudaDad414, line 101:
```suggestion
        oneOf: operation.messages().filterBySend().map(m => m.payload().json())
```
it doesn't work for me without the `.json()` in `dummy` example

src/adapters/mqtt/index.ts
```diff
@@ -42,22 +42,27 @@ class MqttAdapter extends Adapter {
...
-        return this.parsedAsyncAPI.components().securityScheme(secName)
+
+    let userAndPasswordSecurityReq
+    let X509SecurityReq
+
+    const securityRequirements = this.AsyncAPIServer.security().map(e => e.map(e => e.scheme()))
+
+    securityRequirements.forEach(security => {
+      for (const sec of security) {
+        if(sec.type() === 'userPassword') {
```
KhudaDad414, line 53: shouldn't we make this options case-insensitive?

src/adapters/mqtt/index.ts
```diff
@@ -42,22 +42,27 @@ class MqttAdapter extends Adapter {
...
   private getSecurityReqs() {
-    const securityRequirements = (this.AsyncAPIServer.security() || []).map(
-      (sec) => {
-        const secName = Object.keys(sec.json())[0]
-        return this.parsedAsyncAPI.components().securityScheme(secName)
+
+    let userAndPasswordSecurityReq
+    let X509SecurityReq
+
+    const securityRequirements = this.AsyncAPIServer.security().map(e => e.map(e => e.scheme()))
```
KhudaDad414, line 49:
```suggestion
    const securityRequirements = this.AsyncAPIServer.security().map(e => e.all().map(e => e.scheme()))
```

src/adapters/mqtt/index.ts
```diff
@@ -205,8 +209,7 @@ class MqttAdapter extends Adapter {
 
   _send(message: GleeMessage): Promise<void> {
     return new Promise((resolve, reject) => {
-      const operation = this.parsedAsyncAPI.channel(message.channel).subscribe()
-      const binding = operation ? operation.binding('mqtt') : undefined
+      const binding = this.parsedAsyncAPI.channels().get(message.channel).bindings().get('mqtt').value()
```
KhudaDad414, line 212: it gives :
```

x Cannot read properties of undefined (reading 'value')
    at file:///Users/khudadad/Projects/test/glee/dist/adapters/mqtt/index.js:167:103
```
for this line in `dummy` example.

KhudaDad414 (commented):
Left some comments.

src/adapters/mqtt/index.ts
Souvikns: According to docs it should be `userPassword`.

Screenshot 2023-10-13 at 3 06 12 PM

src/adapters/mqtt/index.ts
Souvikns: This is fixed now right, because the `dummy` example is working fine now.

KhudaDad414 (approved):

fmvilas (approved):
LGTM

KhudaDad414 (approved):

PR #551, opened 2023-10-25
feat: add request/reply support

### LINKED ISSUE #544: support address parameters in channels
### Reason/Context
suppose I have the following AsyncAPI spec file in version 3 of the spec:

```yaml
...
servers:
  slack:
    host: wss-primary.slack.com
    protocol: wss
channels:
  main:
    address: "/link/?ticket={ticket}&app_id={appID}"
    parameters:
      ticket:
        description: "your ticket ID. Ticket id can only be used for once."
        location: "$message.payload#/messageId"
        enum: ["a", "b"]
        default: ""
      appID:
        description: "Your slack app ID"
        enum: ["ffff", "ddddd"]
        default: ""
    bindings: 
      ws: {}
    messages:
      reactionAdded:
        $ref: "#/components/messages/reactionAdded"
...
```
it will fail because the [parameters](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.14#parametersObject) are going to be ignored.
### Description
Glee should be able to handle the parameters.
#### If there is a `location`

parse the values from the message's payload or header and pass it to the operation function.
the new glee function type should be:
|Attribute|Description|
|----|----|
|payload|The payload/body of the received event.
|parameters|The parameters of the channel's address.
|headers|The headers/metadata of the received event.
|channel|The name of the channel/topic from which the event was read.
|serverName|The name of the server/broker from which the event was received.

> NOTE: glee should pick up the default value and use it if `location` is missing.

> NOTE: if `enum` is specified, the parameter should be validated.

### LINKED ISSUE #550: Add request/reply support

KhudaDad414:
resolves #550
resolves #544

checklist:
- [x] implement request/reply.
- [x] add channel parameter parsing from payload/headers/asyncapi
- [x] Resolve the double message processing problem.


examples/slack-reaction-listener/functions/reactionAdded.ts
```diff
@@ -0,0 +1,24 @@
+import { GleeFunction } from "@asyncapi/glee"
+enum EVENT_TYPE {
+  EVENTS_API = "events_api",
+  CONNECTED = "hello"
+}
+const myFunction: GleeFunction = async ({ payload }) => {
+  const reaction = payload?.payload?.event?.reaction
+  if (!reaction) return
```
Souvikns, line 8: I am getting this error `Cannot read properties of undefined (reading 'length')` when I am running the example and basically, this error is being thrown from https://github.com/asyncapi/glee/blob/163ff07844b8758e623498dcf2a272dcd16cd509/src/lib/util.ts#L93

Where the `data` parameter is undefined and I think `betterAjvErrors` is not handling it correctly. Also I double checked `reactionAdded` function is returning `undefined`.

KhudaDad414:
@fmvilas @Souvikns a review would be appreciated.

src/adapters/http/client.ts
```diff
@@ -52,30 +53,39 @@ class HttpClientAdapter extends Adapter {
           url = modedAuth.url.href
           query = modedAuth.query
         }
+        if (body && !this.shouldMethodHaveBody(method)) {
+          logWarningMessage(`"${method}" can't have a body. please make sure you are using the correct http method for '${channelName}' channel. ignoring the body...`)
```
fmvilas, line 57:
```suggestion
          logWarningMessage(`"${method}" can't have a body. Please make sure you are using the correct HTTP method for the '${channelName}' channel. Ignoring the body...`)
```

src/adapters/ws/client.ts
```diff
@@ -38,20 +41,26 @@ class WsClientAdapter extends Adapter {
...
-        channel,
+        channel: channelName,
         client: new ws(url, { headers }),
-        binding: this.parsedAsyncAPI.channels().get(channel).bindings().get('ws'),
+        binding: this.parsedAsyncAPI.channels().get(channelName).bindings().get('ws'),
       })
     }
 
     for (const { client, channel } of this.clients) {
+      console.log("openning a connection for channel: ", channel)
```
fmvilas, line 63: Is this console.log intended to stay or just a leftover?

src/lib/adapter.ts
```diff
@@ -54,6 +55,7 @@ class GleeAdapter extends EventEmitter {
     this._parsedAsyncAPI = parsedAsyncAPI
     this._channelNames = this._parsedAsyncAPI.channels().all().map(e => e.id())
     this._channelAddresses = this._parsedAsyncAPI.channels().all().map(c => c.address())
+    this._operationIds = this._parsedAsyncAPI.operations().all().map(o => o.operationId())
```
fmvilas, line 58: In v3, there's no such thing as `operationId` so I'm betting that it will get removed pretty soon from the parser API. I'd instead rely on `.id()`:

```suggestion
    this._operationIds = this._parsedAsyncAPI.operations().all().map(o => o.id())
```

src/lib/functions.ts
```diff
@@ -144,33 +154,64 @@ export async function trigger({
...
     } else {
-      throw err
+      logError(err)
+      return
     }
   }
+}
+
+function createReply(functionReply: GleeFunctionReturnReply, message: GleeMessage, parsedAsyncAPI: AsyncAPIDocumentInterface): GleeMessage {
+  const operation = message.operation
```
fmvilas, line 194:
```suggestion
  const { operation } = message
```

🙂

src/lib/functions.ts
```diff
@@ -144,33 +154,64 @@ export async function trigger({
...
+  const reply = operation.reply()
+  if (!reply) {
+    const warningMsg = `Operation ${operation.id()} doesn't have a reply field. the return result from your function will be ignored.`
```
fmvilas, line 197:
```suggestion
    const warningMsg = `Operation ${operation.id()} doesn't have a reply field. The returned result from your function will be ignored.`
```

src/lib/functions.ts
```diff
@@ -144,33 +154,64 @@ export async function trigger({
...
+  let replyChannel = parsedAsyncAPI.channels().all().filter((c) => c.address() === reply.channel().address())[0]
+  const replyAddress = reply.address()
+  if (replyAddress) {
+    const channelAddress = extractExpressionValueFromMessage(this, replyAddress.location())
+    if (!channelAddress) {
+      throw Error(`cannot parse the ${replyAddress.location()} from your message.`)
+    }
+    const channel = parsedAsyncAPI.allChannels().filter((c) => c.address === channelAddress)[0]
+    if (!channel) {
+      throw Error(`cannot find a channel with the address of "${channelAddress}" in your AsyncAPI file.`)
```
fmvilas, line 211:
```suggestion
      throw Error(`Cannot find a channel with the address "${channelAddress}" in your AsyncAPI file.`)
```

src/lib/functions.ts
fmvilas: Is it true though? I don't think the returned result of the function will be ignored. I mean, we still process `send` if it's there, right?

src/lib/message.ts
```diff
@@ -43,6 +47,7 @@ class GleeMessage extends EventEmitter {
    * @param {Object} [options.headers] Message headers.
    * @param {String} [options.channel] Message channel.
    * @param {String} [options.serverName] The name of the associated AsyncAPI server.
+   * @param {GleeMessage} [options.request] The request that this message is created for as a reply.
```
fmvilas, line 50: > The request that this message is created for as a reply.

What? 😅  Would you mind to elaborate?

fmvilas (changes requested):

src/lib/message.ts
KhudaDad414: In the previouse codebase we were manipulating the message to create a reply. I suggest we create an entirely new message and keep a reference to the parent(request) messge here.
here is a few benifits:
1) by not modefying the message we keep the codebase simpler.
2) The hierarcy of the request/reply will be kept. for example here: 
https://github.com/asyncapi/glee/blob/8e5a1a148927e4fcff07b166285eb4dc60f1e264/examples/slack-reaction-listener/functions/ReceiveFromOpenAI.ts#L4

I am able to trace the request that the reply message was created for.

not sure if this is the best way of doing it though.

src/lib/functions.ts
KhudaDad414: `createReply` will only be called from `reply` not `send`.
I am pretty sure it is going to get ignored but if it should? that's another discussion that we can have. :)
btw I am in favour of ignoring it since I think glee should follow spec as much as possible. no reply in spec? no reply in glee. :|

examples/slack-reaction-listener/asyncapi.yaml
```diff
@@ -0,0 +1,169 @@
...
+    servers:
+      - $ref: "#/servers/OpenAI_HTTPS"
+    address: /v1/chat/completions
+    messages:
+      OpenAIRequest: 
+        $ref: "#/components/messages/OpenAIRequest"
+  SlackEventStream:
+    servers:
+      - $ref: "#/servers/Slack_WebSocket"
+    address: /link/?ticket={ticket}&app_id={app_id}
```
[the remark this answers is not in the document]
KhudaDad414: Will update it to use the bindings.

examples/slack-reaction-listener/asyncapi.yaml
```diff
@@ -0,0 +1,169 @@
...
+      $ref: "#/channels/OpenAICompletion"
+    messages:
+      - $ref: "#/components/messages/OpenAICompletionResponse"
+    reply:
+      channel:
+        $ref: "#/channels/SlackPostMessage"
+  HandleSlackReaction:
+    reply:
+      channel: 
+        $ref: "#/channels/SlackEventStream"
```
derberg, line 87: this feels weird, as your app is reaction listener, and I bet slack websocket server that you listen to do not expect any reply from the app that you're creating. So you receive stream and that is it

src/lib/functions.ts
```diff
@@ -142,35 +152,77 @@ export async function trigger({
...
+function createReplies(functionReply: GleeFunctionReturnReply, message: GleeMessage, parsedAsyncAPI: AsyncAPIDocumentInterface): GleeMessage[] {
+  const operation = message.operation
+  const reply = operation.reply()
+  if (!reply) {
+    const warningMsg = `Operation ${operation.id()} doesn't have a reply field. the return result from your function will be ignored.`
+    logWarningMessage(warningMsg)
+    return
+  }
+
+  let replyChannel = parsedAsyncAPI.channels().all().filter((c) => c.address() === reply.channel().address())[0]
```
fmvilas, line 213: Please provide feedback on https://github.com/asyncapi/parser-api. This is too difficult to read. Maybe we can come up with a simpler way to get this information.

Something like:
```suggestion
  let replyChannel = parsedAsyncAPI.channels().filterByReplyAddress(reply.channel().address())[0]
```

And, IMHO, all collections should have methods like `.first()`, `.last()`, etc. We're doing a lot of `[0]` in many places. This is something that can be reported to Parser API as well. With this, it would look like this:
```suggestion
  let replyChannel = parsedAsyncAPI.channels().filterByReplyAddress(reply.channel().address()).first()
```

fmvilas (changes requested):

examples/slack-reaction-listener/asyncapi.yaml
KhudaDad414: It actually needs a reply. if slack doesn't receive a confirmation it will resend the message multiple times.
https://api.slack.com/apis/connections/socket#acknowledge

KhudaDad414: It actually needs a reply. if slack doesn't receive a confirmation it will resend the message multiple times.
https://api.slack.com/apis/connections/socket#acknowledge

src/lib/functions.ts
KhudaDad414: it makes sense.

https://github.com/asyncapi/parser-api/issues/105
https://github.com/asyncapi/parser-api/issues/104

KhudaDad414:
@fmvilas @Souvikns @fmvilas this PR is ready for review.

examples/slack-reaction-listener/asyncapi.yaml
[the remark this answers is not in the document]
KhudaDad414: https://github.com/asyncapi/glee/issues/591

fmvilas (approved):
LGTM 👍

Souvikns (approved):

PR #579, opened 2023-11-15
fix: use HTTP bindings correctly

### LINKED ISSUE #553: use the correct bindings for http client
#### Describe the bug
in the current codebase we are using the channel binding object. but we do not have channel bindings. https://github.com/asyncapi/bindings/tree/next-major-spec/http

officially we only have operations and messages biddings and glee should parse them from the operation not channel.

KhudaDad414:
**Description**

- [x] Refactor HTTP client code for better maintainablility.
- [x] Refactor HTTP server code for better maintainablility.
- [x] Make sure both client and server use HTTP bindings to validate the requests.
- [x] Make sure the request/reply works with the HTTP protocol.

**Related issue(s)**
Resolves #553


KhudaDad414:
@fmvilas @Souvikns please take a look when you have the time.

examples/http-test/asyncapi.yaml
```diff
@@ -0,0 +1,139 @@
...
+              a:
+                type: string
+              b:
+                type: string
+      payload:
+        type: string
+    empty:
+      payload:
+        type: "null"
+x-remoteServers:
```
fmvilas, line 138: On another PR, not this one, can we get rid of this and move it to the glee.config.js file? It feels weird to see it as part of the AsyncAPI file.

fmvilas (changes requested):

fmvilas (approved):
Approving as my only comment should not block this PR.

Issue #480, opened 2023-07-24
[Docs] Milestones for Creating Tutorial Documentation Guides for Glee


afzal442:
* Related to #431 

**Milestone 1: Topic Preparation**

- [x] Compile a list of important topics related to Glee.
- [x] Identify areas that require guidance from mentors.
- [x] Seek feedback and approval from mentors on the topic list.

**Milestone 2: Brainstorm**

- [x] Brainstorm and discuss the structure of the tutorial documentation with mentors.
- [x] Create a clear and comprehensive overview page for the documentation.
- [x] Outline step-by-step instructions for installation and project setup.

**Milestone 3: Documentation**

- [x] Develop separate sections or guides for each feature of Glee.
- [x] Explain the purpose, usage, and customization options for each feature.
- [x] Incorporate feedback from mentors to improve the documentation.

**Milestone 4: Code Examples and Best Practices**

- [x] Provide practical code examples to demonstrate the usage of Glee's features.
- [x] Incorporate best practices and guidelines for effective utilization of Glee.
- [x] Review and refine code examples based on feedback from mentors.

**Milestone 5: Auto Deployment Guide**

- [x] Create a comprehensive guide for deploying Glee.
- [x] Cover different deployment options and hosting considerations.
- [x] Document specific steps or configurations required for deployment.

**Milestone 6: Interactive Tutorials**

- [x] Develop interactive tutorials using platforms like Killercoda.
- [x] Allow users to experiment with Glee directly within the documentation.
- [x] Ensure a hands-on approach for enhanced learning.

**Milestone 7: Community Engagement**

- [x] Encourage the Glee user community to contribute to the documentation.
- [x] Set up a process for accepting and reviewing contributions.
- [x] Regularly update the documentation based on community submissions.

**Milestone 8: Documentation Release and Maintenance**

- [x] Finalize and publish the tutorial documentation or separate guides
- [x] Make the documentation available on the Glee website or relevant platforms
- [x] Establish a process for ongoing maintenance and updates to keep the documentation current and relevant

### Scope

- [x] Milestone 1: Topic Preparation
- [x] Milestone 2: Brainstorm
- [x] Milestone 3: Documentation
- [x] Milestone 4: Code Examples and Best Practices
- [x] Milestone 5: Auto Deployment Guide
- [x] Milestone 6: Interactive Tutorials
- [x] Milestone 7: Community Engagement
- [ ] Milestone 8: Documentation Release and Maintenance

[...]

Issue #431, opened 2023-05-10
Proposal for having a tutorial document or separate guides for glee. 


Souvikns:
#### Reason/Context

Currently, it is hard to get started with Glee as a new user, we need some kind of documentation that talks about:

- how to install and setup Glee
- achieve what we are trying to do (details on what tweaks we can make)
- how to deploy glee. 

I think it would really benefit from having tutorial documentation for Glee so as to talk about different features Glee has and also as a user manual for new users to learn from, something like [reactjs tutorial](https://legacy.reactjs.org/tutorial/tutorial.html#taking-turns), or we can have a guides sections where we can have setup by setup guides for achieving certain things, and when we have new features we can create guides to showcase them. 

We need to decide if it is something that would be beneficial, and if we decide on this, how should we go about it, should this be on the website or in the Readme, or on wikis? 

#### Specific Goals 
Just defining what we need to document

- We need to document a getting started guide where we help the user set up a glee project and help him get started with the project. 
- Docs that talk about details on different features of glee, like what are different lifecycle events and how to use them, docs on config, etc. 

> pinging @fmvilas @KhudaDad414

rukundob451:
@Souvikns, is this available to be worked on?

rukundob451:
@Souvikns will be waiting in on your response, thank you.

rukundob451:
> @rukundob451 we are thinking of adding this issue to the mentorship program, nothing is final yet give us some time.

I would surely want to be assigned this issues and work on it during the mentorship program.

ConnectBhawna:
Hii @Souvikns I am interested in working in this issue.
Could you please assign this to me. Or firstly may i need to share my approach for the same

maveraw:
Hi @rukundob451, I am willing to contribute to glee, can you tell me where to submit proposal and do we need to contribute also??

AceTheCreator:
Hey Folks, 

To clarify how to submit a proposal, you should contact the project Mentor/s on Slack and send them your submission in the form of a document or ask them how best they'd prefer you submit your proposals.

rukundob451:
> Hey Folks,
> 
> To clarify how to submit a proposal, you should contact the project Mentor/s on Slack and send them your submission in the form of a document or ask them how best they'd prefer you submit your proposals.

Thank you for the communication @AceTheCreator.

PR #580, opened 2023-11-16
docs: adds an interactive tut ref for killercoda env

afzal442:
**Description**

This PR introduces interactive example reference for glee features.

**Related issue(s)**

* Related to #431 
* Part of #480


afzal442:
/rtm

Thanks!

PR #582, opened 2023-11-17
fix: use channel.address instead of id

### LINKED ISSUE #581: Incorrect MQTT Topic Subscription - Uses channel.id() Instead of channel.address()
**Issue Description:**
When defining channels in the AsyncAPI document, Glee seems to be subscribing to MQTT topics based on the `channel.id()` rather than the `channel.address()`. This behavior is inconsistent with the AsyncAPI specification, where the `address` field is intended to specify the MQTT topic.

**AsyncAPI Document Snippet:**
```yaml
channels:
  lightMeasured:
    address: 'light/measured'
    # ... other configurations ...
```

**Expected Behavior:**
Glee should subscribe to the MQTT topic specified in the `address` field of the channel (`light/measured` in this case).

**Actual Behavior:**
Glee is subscribing to a topic derived from `channel.id()` field of the channel (`lightMeasured` in this case).

**Steps to Reproduce:**
1. Define a channel in the AsyncAPI document with a specific `address`.
2. Start the Glee server in development mode.
3. Observe the MQTT topic to which Glee subscribes (different from the specified `address`).
4. Publish a message to the intended topic.
5. Notice that the message is not received by the Glee server.

Souvikns:
**Description**

This small fix updates `channel.address()` instead of `channel.id()` in mqtt.

**Related issue(s)**

Fixes #581

### DIFF
--- a/src/adapters/mqtt/index.ts
+++ b/src/adapters/mqtt/index.ts
@@ -110,7 +110,7 @@ class MqttAdapter extends Adapter {
     this.client.on('close', () => {
       this.emit('close', {
         connection: this.client,
-        channels: this.channelNames,
+        channels: this.channelAddresses,
       })
     })
 
@@ -140,7 +140,6 @@ class MqttAdapter extends Adapter {
   private subscribe(channels: string[]) {
     channels.forEach((channel) => {
       const binding = this.parsedAsyncAPI.channels().get(channel).bindings().get('mqtt')?.value()
-      console.log(binding)
       this.client.subscribe(channel, {
         qos: binding?.qos ? binding.qos : 0,
       }, (err, granted) => {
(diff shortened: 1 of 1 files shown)


KhudaDad414 (commented):
Hey @Souvikns , can you also make sure that when we receive a message, we pass the channel name to the message when it is created not channel address, since glee will verify the channel name.

Souvikns:
Yeah Looks like we are doing it
Screenshot 2023-11-17 at 6 36 23 PM

fmvilas (approved):

KhudaDad414 (changes requested):
When we receive a message and create the message here:

```ts
    return new GleeMessage({
      payload: packet.payload,
      headers,
      channel: packet.topic,
    })
  }
```

  we use the channel address (which is topic) but glee expects to receive the channel id.
  I mean we should change this code to create the message with the channel id not channel address here.
  If you run the current code with an example, it will fail.

KhudaDad414 (approved):
LGTM!

derberg:
@Souvikns you will need to manually update CLI -> https://github.com/asyncapi/cli/blob/master/assets/create-glee-app/templates/tutorial/package.json#L19 as the bumper that we use, is capable to updating just one dependency in given repo, and first on the list that it takes into account is the default one only

PR #598, opened 2023-11-22
feat: support message logging configuration

### LINKED ISSUE #557: provide a flag to disable the logging of inbound and outbound messages
#### Reason/Context
when there are multiple server and long messages are being sent back and forth, the log becomes unreadable due to tremendous amount of message logs. It would be great to have a flag in the config file so user can disable the message logs.

#### Description
I suggest we add an option to disable the logging.
something like:

> glee.config.js

```js
...

glee: {
  logInMessages: true/false
  logOutMessages: true/false
}
...
```

KhudaDad414:
**Description**

- add the config to disable message logs.

**Related issue(s)**
fixes #557


Souvikns (approved):
LGTM 👍🏼

PR #637, opened 2023-12-11
fix: fix some bugs regarding request/reply

KhudaDad414:
**Description**
- validate incoming and outgoing messages including reply messages.
- having a `send` operation is not required when replying.

### DIFF
--- a/src/lib/functions.ts
+++ b/src/lib/functions.ts
@@ -170,19 +170,17 @@ export async function trigger({
     })
 
     functionResult?.reply?.forEach((reply) => {
-      const replyMessages = createReplies(reply, message, parsedAsyncAPI)
-      const hasReplyMessages = replyMessages && replyMessages.length > 0
-      if (!hasReplyMessages) {
+      const replyMessage = createReplies(reply, message, parsedAsyncAPI)
+      if (!replyMessage) {
         return
       }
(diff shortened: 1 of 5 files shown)


Souvikns (approved):

src/lib/functions.ts
```diff
@@ -224,13 +222,5 @@ function createReplies(functionReply: GleeFunctionReturnReply, message: GleeMess
...
-  const sendOperations = replyChannel.operations().filterBySend()
-
-  if (!sendOperations || sendOperations.length === 0) {
-    const warningMsg = `No 'send' operations defined for channel '${replyChannel.id()}'. Ensure your AsyncAPI file defines a 'send' operation for this channel to enable message replies. As a result, no reply will be sent for the channel.`
-    logWarningMessage(warningMsg)
-    return []
-  }
-
-  return sendOperations.map(operation => new GleeMessage({ ...functionReply, channel: replyChannel.id(), request: message, operation, connection: message.connection }))
+  return new GleeMessage({ ...functionReply, channel: replyChannel.id(), request: message, operation, connection: message.connection })
```
fmvilas, line 225: I think we have to create an issue for this. We still need to check but not just on "send" operations but also on replies of a "receive" operation.

PR #658, opened 2023-12-15
fix: fix bindings bug and slack example

KhudaDad414:
**Description**

- add a clear error for when sending a message to HTTP fails.
- fix up the Slack example.


Souvikns (approved):
LGTM 🚀

PR #680, opened 2024-01-09
feat: remove the reply filed support in operations

KhudaDad414:
This PR removes the request/reply pattern support from Glee.

Reason for Removal: Glee uses bindings for its protocols, but the current spec allows replies to different channels, complicating the required bindings for request/reply support in Glee's version 3.

Why Remove Reply Support Entirely?: To avoid confusion, we're removing the reply feature until [AsyncAPI spec issue #1009](https://github.com/asyncapi/spec/issues/1009) is resolved, as 'reply' has different meanings in the spec and Glee framework.


docs/pages/function-lifecycle-events.md
```diff
@@ -37,7 +39,7 @@ export default async function (event) {
 |Attribute|Type|Description|
 |---|---|---|
 |send|array&lt;[OutboundMessage](#anatomy-of-an-outbound-message)&gt;|A list of outbound messages to send when the processing of the inbound event has finished. All clients subscribed to the given channel/topic will receive the message.
-|reply|array&lt;[OutboundMessage](#anatomy-of-an-outbound-message)&gt;|A list of outbound messages to send as a reply when the processing of the inbound event has finished. This is useful when the target of your message is the sender of the inbound event. Note, however, that this only works when you're running Glee as a server. For example, using `reply` when receiving a WebSocket message is fine and the reply will exclusively go to the client that sent the message. However, if you're receiving a message from an MQTT broker, `reply` will work exactly the same way as `send` above, and will send the message to all the clients subscribed to the given channel/topic.
```
afzal442, line 40: interesting, reply is no more. it turned out to be obsolete. thanks for the update

KhudaDad414: it was cool to have it, you didn't have to specify the channel and the server that you sent the message to. but I think it's a source of confusion for now that we release version 1.0. this is something that we need for v2 of glee.

Souvikns (approved):
LGTM 🎉  . I will update the rest of the examples

PR #685, opened 2024-01-14
Moved x-remoteServers Configuration to glee config file

### LINKED ISSUE #601: Move `x-remoteServers` Configuration to `glee.config.(js|ts)`
### Context
Currently, Glee specifies which servers are remote by using the `x-remoteServers` attribute in the AsyncAPI file. It is proposed to move this configuration to the `glee.config.(js|ts)` file, as this is a configuration specific to Glee, rather than to the API itself.

#### Current Implementation
In the current setup, the AsyncAPI file contains the `x-remoteServers` attribute:

```yaml
x-remoteServers:
  - httpbin.org
```

#### Proposed Implementation
The new configuration would be part of the `glee.config.(js|ts)` file, making it more intuitive and aligned with other Glee configurations:

```javascript
  return {
    glee: { // Glee core configurations
      remoteServers: ["httpbin.org"],
      // ... other configurations
    },
    // ... other configurations...
  }
}
```

ayush3160:
**Description**
Changed the logic of fetching x-remoteServers configuration from asyncapi file to fetching from config file.

**Related issue(s)**
Fixes #601


KhudaDad414:
hey @ayush3160 thanks for the contribution.
but I don't think we need to incorporate this feature into Glee now. 

https://github.com/asyncapi/glee/issues/601#issuecomment-1891652040

for now, this will enable us to keep all of the configuration in a central location but since it has some unwanted consequences. I think we should let go of this feature for now.

sorry.

Issue #476, opened 2023-07-10
Glee crashes when body is not proper json in http request


KhudaDad414:
#### Reason/Context

If the the request body does not contain proper json as glee expects, it will crush with the following error:
<img width="897" alt="Screenshot 2023-07-10 at 17 13 00" src="https://github.com/asyncapi/glee/assets/32505158/1e0dc1b1-d624-4968-9670-1d3534c6c8dd">

#### Expected behaviour

1) check if user's HTTP request is documented in AsyncAPI file, if it is not, then show proper message on server console and send a `404` error to the client.

2) Glee should show a proper message instead of crushing if proper json is not passed to by the client.

SrijanShovit:
Hi @KhudaDad414 I would like to take up this issue.
Right now I am trying to setup things.
Thanks!!

SrijanShovit:
Hey I tried setting up, but got confused how to run the project😅

Souvikns:
@SrijanShovit to run the project:
- Fork and clone to project to your local machine
- then install the required dependencies by running `npm i` 
- then build the project by running `npm run build`, once this command is successfully run you will see a `dist` folder in your working directory. 
- Now you can the examples in the [example folder](https://github.com/asyncapi/glee/tree/master/examples)
- or you can create a new example to tests what you want just add the local build as a dependency https://github.com/asyncapi/glee/blob/b0f55a6f83e02cb3004621331fba95968f91f553/examples/dummy/package.json#L13

I hope this helps, if it does not feel free to share which step you are stuck with.

SrijanShovit:
Hi  @KhudaDad414  @Souvikns , I tried to run the project as per the steps told above,but got the following error:
<img width="957" alt="async api" src="https://github.com/asyncapi/glee/assets/86592220/c9f02cb0-279e-48c8-b2a9-23c24d628069">

I tried understanding about the error, but as I am new , I couldn't.
Could you please guide about it and also may I have an overview of flow or architecture of this project?
Thanks!!

KhudaDad414:
@SrijanShovit the error should be gone now. please delete you `.glee` folder and update your glee to the latest version. after that run `npm run dev` and it should be fine.

Souvikns:
Is anyone working on this, should we put this issue for hacktoberfest?

AnishKacham:
Hi. I can work on this if no one's assigned 😅

KhudaDad414:
[...]

PR #696, opened 2024-01-22
fix: updating import for next/env

Souvikns:
**Description**
Screenshot 2024-01-22 at 2 59 56 PM

Updating the import `@next/env` import as it is a commonJS module. 

**Related issue(s)**

See also #476


KhudaDad414:
@Souvikns it has been fixed here: https://github.com/asyncapi/glee/pull/691

PR #712, opened 2024-01-30
fix: removes combination of files options

afzal442:
**Description**
Due to some caveats, selection of combination of filenames in the pipeline job might confuse the github workflow. Or might be this `rm -r ./pages/docs/tools/glee`

**Related issue(s)**


afzal442:
@KhudaDad414 your recent changes on docs didn't reflect on website repo due to workflow error while it ran successfully. I wondered why it didn't catch that.

.github/workflows/update-docs-in-website.yml
KhudaDad414, line 1: Hey @afzal442, can you configure [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) on this as well? so we can run this workflow manually in case something like this happens again?

KhudaDad414 (approved):

PR #737, opened 2024-03-12
fix: send messages to the address of the channel

### LINKED ISSUE #721: MQTT channel seems to be using wrong topic when sending message
#### Describe the bug
I'm working with an MQTT protocol (asyncapi 3.0.0). If I have a channel called 'hello', with an address that is set to 'test/testing/setParamResult', and I use glee to send a message to the channel, it seems like it is published on the MQTT topic 'hello', instead of what I set the address to. Is this expected? It does seem to use the address to determine which topics to subscribe to, so I would have expected it to send messages there as well.

#### How to Reproduce
Create MQTT API spec and create channel in API spec:
```
channels:
  setParamResult:
    address: 'test/testing/setParamResult'
    messages:
      payload:
        setParamResult:
          type: string
```

Then use glee to send a message on this channel. Message is sent to MQTT topic setParamResult instead of 'test/testing/setParamResult'
#### Expected behavior
Message is sent to topic 'test/testing/setParamResult' instead.

KhudaDad414:
**Description**

- fix the bug when Glee sends the message to the channel name instead of the channel address in MQTT.
- show a warning when the channel name is invalid when we send a message.
- show a warning when there is no send operation defined in the AsyncAPI file.

**Related issue(s)**
fixes #721

### DIFF
--- a/src/adapters/mqtt/index.ts
+++ b/src/adapters/mqtt/index.ts
@@ -248,14 +248,11 @@ class MqttAdapter extends Adapter {
 
   _send(message: GleeMessage): Promise<void> {
     return new Promise((resolve, reject) => {
-      const binding = this.parsedAsyncAPI
-        .channels()
-        .get(message.channel)
-        .bindings()
-        .get('mqtt')
-        ?.value()
+      const channel = this.parsedAsyncAPI.channels().get(message.channel)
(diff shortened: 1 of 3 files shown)


Souvikns (approved):
LGTM 👍🏼

PR #769, opened 2024-03-28
fix: error when trying to access the Websocket API if pathname contains trailing slashes.

### LINKED ISSUE #713: but this path is not defined in your AsyncAPI file - Caused by trailing slash
# Overview
Currently, attempts to access the Websocket API, including trailing slashes, result in an error `but this path is not defined in your AsyncAPI file`.
I believe that resolving this will make the library easier to use.

# Solution
Make the following changes to _extractPathname.
If it looks good, assign me!
```typescript
private _extractPathname(req: IncomingMessage) {
  const serverUrl = new URL(this.serverUrlExpanded)
  const { pathname } = new URL(req.url, serverUrl)

  if (!pathname) return '/'

  if (pathname.endsWith('/')) {
    return pathname.substring(0, pathname.length - 1)
  }

  return pathname
}
```
https://github.com/asyncapi/glee/blob/be4a2a61909206561d0487e05e44ff03bb366746/src/adapters/ws/server.ts#L183-L187

# Append
I was careless.
That you can also set an address in `asyncapi.yaml` that includes a trailing slash.
Perhaps we need to remove the trailing slash in the channels method of `@asyncapi/parser` for unification, and also in the tilde method of utils.ts if we want to make it more correct.
https://github.com/asyncapi/parser-js/blob/9742b8145020f43db598fe394702fbab5a9c868e/src/models/v3/asyncapi.ts#L61-L67

https://github.com/asyncapi/parser-js/blob/9742b8145020f43db598fe394702fbab5a9c868e/src/utils.ts#L113-L121

If the impact is too large, I would suggest changing the _getChannel method as follows
```typescript
private _getChannel(req: IncomingMessage) {
  const pathName = this._extractPathname(req)
  return this.parsedAsyncAPI.channels().all().filter(channel => {
    let address = channel.address()
    if(address.endsWith('/')) address = address.substring(0, address.length - 1)

    return address === pathName
  })[0]
}
```
https://github.com/asyncapi/glee/blob/be4a2a61909206561d0487e05e44ff03bb366746/src/adapters/ws/server.ts#L95-L98

NaCl5alt:
**Description**

- Fixed so that connections can be made even if there is a trailing slash when accessing the WebSocket API.

**Related issue(s)**
Fixes #713

### DIFF
--- a/src/adapters/ws/server.ts
+++ b/src/adapters/ws/server.ts
@@ -94,7 +94,12 @@ class WebSocketsAdapter extends Adapter {
 
   private _getChannel(req: IncomingMessage) {
     const pathName = this._extractPathname(req)
-    return this.parsedAsyncAPI.channels().all().filter(channel => channel.address() === pathName)[0]
+    return this.parsedAsyncAPI.channels().all().filter(channel => {
+      let address = channel.address()
+      if (address.endsWith('/')) address = address.substring(0, address.length - 1)
+
+      return address === pathName
+    })[0]
   }
 
   private _endRequest(code: number, message: string, socket: Duplex) {
@@ -183,7 +188,14 @@ class WebSocketsAdapter extends Adapter {
   private _extractPathname(req: IncomingMessage) {
     const serverUrl = new URL(this.serverUrlExpanded)
     const { pathname } = new URL(req.url, serverUrl)
-    return pathname || '/'
+
+    if (!pathname) return '/'
+
+    if (pathname.endsWith('/')) {
+      return pathname.substring(0, pathname.length - 1)
+    }
+
+    return pathname
   }
 
 


Souvikns (approved):
LGTM 🚀

src/adapters/ws/server.ts
```diff
@@ -94,7 +94,12 @@ class WebSocketsAdapter extends Adapter {
 
   private _getChannel(req: IncomingMessage) {
     const pathName = this._extractPathname(req)
-    return this.parsedAsyncAPI.channels().all().filter(channel => channel.address() === pathName)[0]
+    return this.parsedAsyncAPI.channels().all().filter(channel => {
+      let address = channel.address()
+      if (address.endsWith('/')) address = address.substring(0, address.length - 1)
```
KhudaDad414, line 99: @NaCl5alt what if the address is `/`?

